PR #82, opened 2023-07-07
Update documentation to reflect the removal of IControl on Avalonia 11.

cotti:
Hello. 

The docs for v11 retained references to IControl, which was removed. So this quick PR patches those.


maxkatz6 (approved):
Thanks!

PR #79, opened 2023-07-04
Initialize Simplified Chinese support. 

rabbitism:
This PR initializes the Simplified Chinese documentation support. 
Actual translations will come after this.

locally tested with 
```bash
npm run start -- --locale zh-Hans
```

PR #84, opened 2023-07-07
Revert "Initialize Simplified Chinese support. "

jmacato:
Reverts AvaloniaUI/avalonia-docs#79


jmacato:
@rabbitism Seems like duplicated pages are not recommended because it might ruin the searchability of the docs site. It would be better to have fully translated pages first instead of fully copying the old english content over to the chinese one as per @kekekeks

rabbitism:
> @rabbitism Seems like duplicated pages are not recommended because it might ruin the searchability of the docs site. It would be better to have fully translated pages first instead of fully copying the old english content over to the chinese one as per @kekekeks 

Yes, I'm ok with reverting, we will do translation in our fork first. However, It is possible for Chinese community because we have quite some volunteers working on this, but It's gonna be hard for other languages. 
Now people have following options:
1. A translator has to finish all pages to contribute, that is quite discouraging.
2. Only add translated pages to the folder, which forces every translator to take care of folder structure and menu json config, which is a very bad experience and i believe there will be many merge conflicts.
3. Leave pages empty if not translated, which means user have no fallback and have to switch back to english to check. 

Anyway I respect your decision, and we ideally can finish all in a week. But please consider how it is going to work with other language translators. 



Anyway I'm

maxkatz6:
I wonder if in translated sidebar it's possible to add files from non-translated version. I.e. keeping a single file/page in both sidebars.

jmacato:
@rabbitism sorry if i wasnt clear on the wording, what i meant was we dont have to move all the english pages over all at one and instead just incrementally add the translated ones as the translation is going on :)

PR #85, opened 2023-07-07
Fix sample so that it compiles

Nippius:
When copy/pasting the sample code, the code will not compile due to a typo in the namespace declaration


maxkatz6 (approved):
Thanks!

PR #97, opened 2023-07-11
Update textbox.md to include `revealPasswordButton` style class

DamianSuess:
This PR provides documentation and sample code for using the `revealPasswordButton` style class on the TextBox user control.

### DIFF
--- a/versioned_docs/version-0.10.x/controls/textbox.md
+++ b/versioned_docs/version-0.10.x/controls/textbox.md
@@ -60,7 +60,7 @@ produces the following output in **Windows 10** when text is input
     <img src="/img/controls/textbox/textbox_password.png"  />
   </div>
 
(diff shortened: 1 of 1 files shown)


maxkatz6:
Thanks! The same change should be applied to the textbox.md file for 11.0 version. Also, this class might or might not be supported by third party themes, but is supported by fluent one.

DamianSuess:
Appreciate the feedback as always @maxkatz6. Good callout on the Fluent theme. I'll make the updates and get it pushed up

DamianSuess:
@maxkatz6, for the v11 docs, is that located in the path: `/avalonia-docs/docs/reference/controls/detailed-reference/textbox.md`

maxkatz6 (approved):
Thanks!

maxkatz6:
@DamianSuess oh, haven't noticed your last comment. Yes, that's right, the second textbox file is located there.

PR #99, opened 2023-07-13
Fixed ToDo List App tutorial docs

n20amaral:
I tried to follow the ToDo List App tutorial, but I struggled with some issues. I was using the latest version 11.0.0 with Visual Studio Extension. After some effort, because I never used AvaloniaUI and/or WPF before, I could make it work and decided to contribute and update this tutorial.

I think the most of my problems were related with the new .csproj file generation that now adds this property `    <AvaloniaUseCompiledBindingsByDefault>true</AvaloniaUseCompiledBindingsByDefault>`.

I'm not sure if this is a valid contribution, but at least if someone new to AvaloniaUI follow the tutorial with these changes, everything will work.

I'm here to help, please be free to email me if you need.

PS: Sorry for the generic commit titles, but there is information in the commit description.


maxkatz6 (approved):
Went through these changes couple of times and it looks good to me.
Thanks!

PR #100, opened 2023-07-14
fix Button Controls sidebar menu

DmitryZhelnin:
Before

изображение

After

изображение

Should `Toggle Button` and `Toggle Split Button` pages be moved to `buttons` folder from `detailed-reference` folder?


maxkatz6:
[sidebars.js](https://github.com/AvaloniaUI/avalonia-docs/blob/main/sidebars.js) file might need to be adjusted too.

> Should Toggle Button and Toggle Split Button pages be moved to buttons folder from detailed-reference folder?

Yes, it would be nice!

DmitryZhelnin:
> [sidebars.js](https://github.com/AvaloniaUI/avalonia-docs/blob/main/sidebars.js) file might need to be adjusted too.

`Reference` section is autogenerated, no changes needed in this case.

> Yes, it would be nice!

Done.

maxkatz6 (approved):
Thanks!

PR #149, opened 2023-08-26
Update return-from-dialog.md

Philippe-Laval:
We edit the axaml file


maxkatz6:
Thank you for finding and resolving these problems!

PR #171, opened 2023-09-11
Corrected misspelled element name (zh version)

Arhell:
ref https://github.com/AvaloniaUI/avalonia-docs/pull/167/files


Arhell (commented):
@maxkatz6

PR #178, opened 2023-09-14
Update dealing-with-platforms.md

timunie:
commented out Avalonia.Essentials as long as they are not production ready. 

/cc @grokys @MikeCodesDotNET


maxkatz6 (approved):

MikeCodesDotNET:
Not a fan. We could have instead said "This is a work in progress" and pointed people towards it as a reference.

PR #179, opened 2023-09-14
Corrected (zh version)

Arhell:
ref
https://github.com/AvaloniaUI/avalonia-docs/pull/175/files
https://github.com/AvaloniaUI/avalonia-docs/pull/174/files


Arhell (commented):
@maxkatz6 cc

Issue #166, opened 2023-09-05
Image convention cleanup


stevemonaco:
As per the readme, files should be stored like the following:

```
- Use lowercase for file and folder names and dashes - as separators.
  For example:
  - /docs/getting-started/ide-extensions.md
  - /docs/how-to/use-google-fonts.md
```

However, there are many images that are not. One example is `/img/data-binding/data-validation/CustomValidationTooltip.png`.

I propose the following goals, which I'm willing to do myself. Most of this needs to happen at once, but I could do the new styling + fixups, then the rename + fixups across two PRs. I'm also uncertain as to how shared the images are with v0.10 as many are within `/img/gitbook-import/assets` and whether sharing is even desirable.

- [X] Introduce image styling with broken link detection (below)
- [X] Migrate v11 Markdown docs to new image styling
- [X] Migrate v0.10.x Markdown docs to new image styling
- [ ] Copy all externally referenced images into the docs repo
- [ ] Rename all image files to match convention with meaningful names and fix up v11 / v0.10.x docs

# New Image Styling

The current styling approach uses `<img src="/img/..." />` directly. This doesn't allow Docusaurus to automatically detect broken images and throw an error when trying to display. This happens automatically if you use the markdown syntax or `img src={require(...)}` syntax. I cannot figure out how to center the markdown (may require writing a plugin which is too much maintenance), so I suggest the second option as follows:

1. Add to custom.css:
```css
img.center {
  display: block;
  margin-left: auto;
  margin-right: auto;
}
```
2. Replace all document image links like this (centering is optional):

> ~~<img className="center" src={require('/img/animations/page-transitions/transitioning-content-control-03.webp').default} alt="test" />~~

Edit: Specifying external resources at the top allows better management, so should be preferred.

```html
import CustomPageTransitionImage from '/img/animations/page-transitions/transitioning-content-control-03.webp';
<img className="center" src={CustomPageTransitionImage} alt="test" />
```

Example CLI error (`npx docusaurus start`) when an image doesn't exist. There's also a giant red popup that shows when the site is loaded (locally, I assume while deployed, too).
```
Module not found: Error: Can't resolve '/img/fake-image.png' in 'C:\Users\Steve\source\repos\avalonia-docs\docs\guides\graphics-and-animation'
```

[...]

PR #182, opened 2023-09-16
Migrate v11 docs to image styling with broken link detection

stevemonaco:
From https://github.com/AvaloniaUI/avalonia-docs/issues/166:

- [X] Introduce image styling with broken link detection
- [X] Migrate v11 Markdown docs to new image styling

All English pages were checked manually while migrating. Chinese pages were diff'd manually with help from Rider's Compare Directories. I loaded most pages manually and seemed fine. `npm run build` detected no image issues for both locales. Fixed a small handful of typos and broken formatting, but tried to keep that to a minimum. Fixed approximately 19 broken images (on each locale).

Tested to work with JavaScript disabled. Build and run via `npm run serve`, not in dev mode.

Will continue with v0.10.x if this is merged.

I used a verbose naming convention of `ConceptDiagram` and `ConceptScreenshot` to show a division between images made by a diagram tool, a screenshot that can't be rendered by Avalonia (eg. a VS screenshot), and an image that conceivably could be produced by the source code and `RenderTargetBitmap`. I didn't succeed on the third point at all, especially due to live screen captures. I plan to make another naming pass at the same time as the image files are renamed/reorganized, so feedback on naming before then would help.


maxkatz6 (approved):
It does feel verbose this way for sure, but it's an improvement anyway:

1. We had lots of problems with broken links to images, and this PR solves this problem.
2. It's still can be compiled into a static page, and doesn't require JS in runtime.

PR #194, opened 2023-10-06
Update process-a-new-item.md

Basicconstruction:
The sample code does not compile or execute properly in version 11.0.04 (cannot even be checked by the editor)

### DIFF
--- a/docs/tutorials/todo-list-app/process-a-new-item.md
+++ b/docs/tutorials/todo-list-app/process-a-new-item.md
@@ -50,14 +50,15 @@ namespace ToDoList.ViewModels
                 addItemViewModel.OkCommand,
                 addItemViewModel.CancelCommand.Select(_ => (ToDoItem?)null))
                 .Take(1)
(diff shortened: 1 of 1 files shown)


docs/tutorials/todo-list-app/process-a-new-item.md
```diff
@@ -50,14 +50,15 @@ namespace ToDoList.ViewModels
                 addItemViewModel.OkCommand,
                 addItemViewModel.CancelCommand.Select(_ => (ToDoItem?)null))
                 .Take(1)
-                .Subscribe(newItem =>
+                .Subscribe(Observer.Create<ToDoItem?>(newItem =>
```
maxkatz6, line 53: For ReactiveUI there is no difference between 0.10 and 11.0 here. In both cases you need to add "using System;".

So, it might need to be added here instead.

Basicconstruction:
it works when i use 
```
using System;
```

PR #198, opened 2023-10-11
Update jetbrains-rider-setup.md - Added missing step to add plugin re…

mladenb:
…po in Rider

### DIFF
--- a/versioned_docs/version-0.10.x/getting-started/jetbrains-rider-setup.md
+++ b/versioned_docs/version-0.10.x/getting-started/jetbrains-rider-setup.md
@@ -55,6 +55,8 @@ A new Preferences Screen will open up. Click the `Settings` icon as shown and se
 
 <img className="center" src={RiderConfigurePluginsScreenshot} alt="Button group" />
 
+Click the `+` icon and enter the URL `https://plugins.jetbrains.com/plugins/dev/14839`then click `OK`.
+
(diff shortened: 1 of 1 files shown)


maxkatz6:
I am pretty sure this plugin was supposed to be discoverable even without a special feed.
cc @ForNeVeR

ForNeVeR:
@maxkatz6 is correct, you are not supposed to enter any additional URLs.

The instruction here is to get a "Dev" build of the plugin (which aren't intended for the general public, since I release new bug fixes and features pretty quick anyway after their merge to `main`).

A normal, "stable" version of the plugin is supposed to be available without adding of any URLs.

mladenb:
I couldn't find this plugin in my Rider until after I added that url.

ForNeVeR:
Please add some details. What version of Rider do you use?

Here's a support matrix for the stable plugin versions.
image

Pretty much every modern Rider version should be able to download the plugin (though the latest 1.2.0 with support for Rider 233 was only released recently).

mladenb:
Hi, yeah, that could be the culprit. Here is my Rider version:
```
JetBrains Rider 2021.2.3
Build #RD-212.5712.52, built on December 23, 2021
```

ForNeVeR:
That's a very old one, and probably I never released a stable version of the plugin for it (but there were some earlier dev versions compatible: that's why you can use them now).

There are no plans to support older versions of Rider. But feel free to use the older dev builds that are available in the said channel; I have no plans on removing these versions either.

My opinion is that it's not worth adding to the documentation.

maxkatz6:
Closing in favor of #217 
Thank you @mladenb !

PR #221, opened 2023-11-02
Update listbox.md

StayMeaty:
Binding the Items property of the ListBox leads to an error because there is no setter or adder for the Items property in Avalonia 11 anymore. The recommended way is to use the ItemsSource property instead.


StayMeaty:
nvm the documentation i relied on isn't avalonia 11

PR #226, opened 2023-11-05
Add more details for built-in control types

robloo:
There are really 4 types of controls in a XAML framework:

 1. Drawn Controls
 2. Layout Controls (no appearance)
 3. User Controls
 4. Templated Controls
     * Fully Customizable (Button)
     * Partially Customizable (DataGrid)

This is a categorization I've been thinking about for a few years now (layout controls with no appearance is a new one based on Avalonia's existing docs). I think it's useful for devs to be aware of this information and I haven't seen it anywhere else. Usually people just say there are 'lookless' templated controls in XAML but in practice there are caveats and differences to be aware of that limit what is possible for an app.


sidebars.js
```diff
@@ -163,7 +163,7 @@ const sidebars = {
                     'id': 'basics/user-interface/controls/creating-controls/index'
                   },
                   'items': [
-                    'basics/user-interface/controls/creating-controls/choosing-a-custom-contro-type',
+                    'basics/user-interface/controls/creating-controls/choosing-a-custom-control-type',
```
maxkatz6, line 166: If any link was changed, we should also update redirects file. Just to keep old links work fine:
https://github.com/AvaloniaUI/avalonia-docs/blob/b88a7ee4ff05bd6b7f8699d198844ff47bada1a3/redirects.js

robloo: Ah, yes, good catch

robloo: Should be fixed

robloo: According to the test failure the link I redirected to was invalid:

```
        {
            from: '/docs/basics/user-interface/controls/creating-controls/choosing-a-custom-contro-type',
            to: '/docs/basics/user-interface/controls/creating-controls/choosing-a-custom-control-type',
        },
```

Is it correct to include 'next' in the path like `/docs/next/basics`? If so, I'm not sure how the other links work. None of the others had it.

maxkatz6: Other redirects were for the old documentation (i.e. 0.10). 
11.0 documentation still uses "next" prefix, so it should be included, yes.

Although, ideally we should remove this prefix and adjust all of these redirects.

robloo: I see. Yes I think the current docs need to remove the 'next' prefix... which would create a redirect for everything.

The 0.10.x docs are the ones that need a version if anything. So it's setup backwards of what I would expect.

robloo: Should be fixed, let's see if tests pass.

robloo:
@maxkatz6 I think I fixed all the errors on my end. It seems there is an issue with the build system now. How would you like me to proceed with this PR?

maxkatz6 (approved):

maxkatz6:
@robloo just fixed main branch. Should be good now.

PR #227, opened 2023-11-05
Add PointerPressed event usage warning to Button

robloo:
See discussion here: https://github.com/AvaloniaUI/Avalonia/discussions/10673


maxkatz6:
Probably, adding an addition note on the events documentation page would also help. Since it's not only about Button, but it's generally how events are handled in XAML frameworks.

maxkatz6 (approved):

robloo:
> Probably, adding an addition note on the events documentation page would also help. Since it's not only about Button, but it's generally how events are handled in XAML frameworks.

Yea, that's probably a good idea. Not volunteering for it quite yet myself though.

PR #230, opened 2023-11-05
Add Initial Colorpicker Docs

robloo:
I copied over the original PR from the old repo, cleaned it up a bit and added a ColorView page. I think this is good enough for now considering no other docs exist yet.

It can be improved going forward.

Please review for technical correctness: images are in the correct location, links are correct, etc.


docs/reference/controls/colorpicker/README.md
```diff
@@ -0,0 +1,113 @@
...
+Several pre-defined color palettes implementing the `IColorPalette` interface are provided. Instances of these palettes may be set to the `Palette` property of a `ColorView` or `ColorPicker`.
+
+<table>
+  <tr>
+    <th>Palette</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>
+      <img src="/img/controls/colorpicker/color-palette-fluent.png" alt="Fluent Color Palette" width="300" />
```
maxkatz6, line 79: For images and links, we now use JS import syntax, so build tool can validate path in compile time. See https://github.com/AvaloniaUI/avalonia-docs/blob/main/docs/reference/controls/buttons/button.md?plain=1#L5

Also, we were slowly moving from HTML to Markdown on some other pages. But it's not a bit deal.

docs/reference/controls/colorpicker/README.md
```diff
@@ -0,0 +1,113 @@
+# ColorPicker
```
maxkatz6, line 1: File header also can be added, see the same button page as a reference. But that's optional with this docs engine.

More importantly, `builtin-controls.md` file should be updated to include ColorPicker as well. 
It seems like we don't need to update sidebar.js for this tho.

docs/reference/controls/colorpicker/README.md
robloo: Images in markdown tables may not always be supported. I think I ran into an issue in the past which required this. I'll use the standard markdown format though and we'll see if it works here.

robloo: Made both changes

docs/reference/controls/colorpicker/README.md
robloo: Also made both recommended changes

maxkatz6 (approved):
Thanks!

docs/reference/controls/colorpicker/README.md
robloo: @maxkatz6 In the published docs only the first color palette image is shown. This type of problem is why the table was HTML.

Is there a way to fix this or should I revert back to HTML here?

robloo: I'm reverting back to HTML here: https://github.com/AvaloniaUI/avalonia-docs/pull/238

PR #238, opened 2023-11-12
Revert "Switch from HTML to markdown table syntax"

robloo:
Follow-up to:
 * https://github.com/AvaloniaUI/avalonia-docs/pull/230

This reverts commit ed3c5407f47fa0dee070a912702e507cffebe40e.

As discussed here: https://github.com/AvaloniaUI/avalonia-docs/pull/230#discussion_r1382671308, Markdown doesn't seem to like images in a table. This reverts back to HTML in an attempt to fix this:

image


maxkatz6:
I don't understand why it doesn't, but the old markdown should be properly transformed into html as well.
Maybe `Example banner` markdown syntax would work better (assuming it also does compile time checks). But let's keep html for now.

maxkatz6 (approved):

PR #239, opened 2023-11-12
Add global resource lookup from code section

robloo:
@timunie I update the section you were working on to include a bit more info for those coming from other XAML UI frameworks. Accessing resources from code is a little tricky in Avalonia (and honestly it needs to be better). Discussion was here:

 * https://github.com/AvaloniaUI/Avalonia/issues/6021

### DIFF
--- a/docs/guides/styles-and-resources/resources.md
+++ b/docs/guides/styles-and-resources/resources.md
@@ -250,7 +250,7 @@ The search moves upwards in the logical control tree, behaving at each level in 
 
 ## Consuming Resources from code
 
-Avalonia provides different options to access Resources from code. 
+_Avalonia UI_ provides different options to access Resources from code. 
 
 :::NOTE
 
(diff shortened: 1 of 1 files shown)


docs/guides/styles-and-resources/resources.md
```diff
@@ -284,10 +284,54 @@ public MainWindow()
...
     myTextBlock.Bind(TextBlock.TextProperty, Resources.GetResourceObservable("TheKey"));
 
     // this will update myTextBlock.Text via the bound observable
     this.Resources["TheKey"] = "Hello from code behind"; 
 }
 ```
+
+### Global Resource Lookup
+
+In other XAML UI frameworks such as WPF and WinUI, resource lookup could also be done using an indexer and code such as `Application.Current.Resources["TheKey"]` (in addition to `FindResource()` or `TryFindResource()`). Resource lookup using the indexer at the application scope would search globally and recursively in all resource dictionaries. You could almost always find a resource using the indexer. _Avalonia UI_ does not support this and will only search the current resource dictionary when using an indexer. Additionally, _Avalonia UI_ may store resources in both resource dictionaries or style resource dictionaries so it's common to be required to search both. For global, recursive resource lookup at the application scope similar to other XAML UI frameworks code such as below may be used.
```
maxkatz6, line 297: But this method is not necessary. Existing TryFindResource should already cover this case. Or am I missing something?

robloo: TryFindResource() won't search BOTH the resources and style resources (these never should have been separated IMO). It is one or the other depending on where you call it.

Additionally, the indexer functionality is completely different from WPF/WinUI which this section explains as well. The new method reproduces exactly what the old WPF indexer resource lookup logic does. Explained further in #6021 and https://github.com/AvaloniaUI/Avalonia/discussions/6019.

maxkatz6: > TryFindResource() won't search BOTH the resources and style resources

It will. https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Application.cs#L215-L216
Controls/Window should also have the same behavior.
Keep in mind, that Fluent/SimpleTheme are styles, that contain resources and it can be reached with TryFindResource 

> Additionally, the indexer functionality is completely different

Yes, that's another problem, that should be mentioned. But TryFindResource specifically works.

robloo: > Keep in mind, that Fluent/SimpleTheme are styles, that contain resources and it can be reached with TryFindResource

I'll test again but I know for a fact I had to add those lines of code iterating through Style resources to get FluentAvalonia resource lookup to work. So unless something has changed in the last few months I'm not sure I agree that it works.

robloo: @maxkatz6 Note I'm not the only one that has found issues in this area: https://github.com/amwx/FluentAvalonia/discussions/378#discussioncomment-6030956

robloo: @maxkatz6 I confirmed again that Styles does need to be searched to pick up resources such as `OverlayCornerRadius` with FluentAvalonia. This indicates that the code snippet I provided is correct.

If you would like I can move all of this to the WPF section: get-started/wpf/styling

robloo:
@maxkatz6 @timunie Should I close this?

timunie:
@robloo I have mixed feelings here and let the team decide. Thx for remembering me anyways

maxkatz6:
(Try)FindResource works with both Control.Resources and Control.Styles here. I still don't understand where is the issue.

[av.zip](https://github.com/AvaloniaUI/avalonia-docs/files/13712317/av.zip)

It works specifically due to:
https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Application.cs#L212-L217 
https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Base/Styling/Styles.cs#L118-L135
https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Base/StyledElement.cs#L453-L458

I assume original confusion goes from the `App.Resources["Name"]` usage, which will ignore styles. Which is exactly the confusing in https://github.com/AvaloniaUI/Avalonia/issues/6021. But as I said, there is no need to yet another "find resource" method.

docs/guides/styles-and-resources/resources.md
maxkatz6: > these never should have been separated IMO

I can agree that it shouldn't be allowed to add resources to the Style. I.e. remove Style.Resources collection completely. But it's couple of years too late for that.

robloo:
@maxkatz6 I wouldn't say there is a confusion here at this point. It simply does not work on my end in a real app and the helper method discussed is required. I checked it again to confirm and this isn't just me either: others using FluentAvalonia are seeing it too. That likely means there is a bug some place. It's not worth our time to discuss this further for documentation so I'll close this. The work-around will remain in my code.

Separately, it would be good to align the indexer lookup implementation to match WPF.

PR #246, opened 2023-11-19
updates to test drive

IvanRainbolt:
updates to the get started - test drive section to match current templates and conventions.


maxkatz6 (approved):
Thank you!

PR #261, opened 2023-11-30
Improve example in accessing-the-ui-thread.md

blue00:
I think the current example is not the best fit to demonstrate how to call the ui thread from another thread. Long running work can simply be solved with await Task.Run(() =>{...}).

Also I suspect that the current example description is not correct. It says "Notice that because the long running task is executed on its own thread, the UI does not lose responsiveness." But the LongRunningTask() Function is actually running on the UIThread. It's only not blocking the ui thread because it uses Task.Delay instead of Thread.Sleep.

Thats why I created a new sample with a usecase I find more fitting. I hope that it is helpfull.

### DIFF
--- a/docs/guides/development-guides/accessing-the-ui-thread.md
+++ b/docs/guides/development-guides/accessing-the-ui-thread.md
@@ -34,75 +34,52 @@ For the possible values of the `DispatcherPriority` enumeration, see [here](http
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


Arhell (approved):

blue00:
Sorry, I thought auto-merging main into this branch wouldn't undo the review.

maxkatz6 (approved):

PR #304, opened 2024-01-05
change: slider binding

rtazaki:


docs/reference/controls/slider.md
```diff
@@ -38,6 +39,45 @@ The slider looks like this on Windows:
...
+```cs
+using System.Windows.Input;
+using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+
+namespace AvaloniaControls.ViewModels;
+
+public class MainViewModel : ViewModelBase
+{
+    [Reactive]
```
maxkatz6, line 64: ReactiveUI.Fody is a bit too specific, and we haven't really mentioned it before in the documentation. We should either omit VM impl details or add at least some comments about Fody.

maxkatz6 (approved):

docs/reference/controls/slider.md
rtazaki: I would like to document a sliderbar example. 
I'd like to avoid relying on a particular framework, but then I won't be able to provide useful examples of how to do the binding.
Therefore, I had to explain it in a little more detail.

PR #327, opened 2024-01-23
Deep Dive on custom ItemsPanel

### LINKED ISSUE #326: Docs about ItemsPanelTemplate usage
@cerobe you could add a page under docs > DeepDrives and describe how to use a custom ItemsPanel For ItemsControl and friends. From the ItemsControl-page please add a link to that docs.

_Originally posted by @timunie in https://github.com/AvaloniaUI/Avalonia/discussions/14296#discussioncomment-8219936_

cerobe:
- Created a new Deep dive page for the customisation of ItemsPanel
- Linked (two ways) the build-in controls that have ItemsPanel
- Modified menu to include the new page

fixes #326


docs/reference/controls/itemscontrol.md
```diff
@@ -17,9 +18,34 @@ To see the full list of _Avalonia UI_ built-in repeating data controls, see [her
...
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>ItemsSource</code></td>
+      <td>The bound collection that is used as the data source for the control.</td>
+    </tr>
+    <tr>
+      <td><code>ItemsControl.ItemTemplate</code></td>
+      <td>Attached property element to contain the data template for an individual item.</td>
```
thevortexcloud, line 35: I know you did you not write this, but the item template property is not an attached property. I am not sure why it says this in the first place. Attached properties are a very specific thing.

cerobe: no problem, what would be the correct description? Should I just remove the 2 leading words "Attached property"

docs/reference/controls/itemscontrol.md
```diff
@@ -17,9 +18,34 @@ To see the full list of _Avalonia UI_ built-in repeating data controls, see [her
...
+    </tr>
+    <tr>
+      <td><code>ItemsControl.ItemPanel</code></td>
+      <td>Parent panel to display the items in. By default, this is a StackPanel</td>
```
thevortexcloud, line 39: I would suggest rewording this to be less ambiguous. "Parent panel" might confuse people and think it's somehow changing the parent of the item control itself. Something like "The container panel to place items in" would work better.

docs/reference/controls/itemscontrol.md
```diff
@@ -17,9 +18,34 @@ To see the full list of _Avalonia UI_ built-in repeating data controls, see [her
...
+    </tr>
+    <tr>
+      <td><code>ItemsControl.Styles</code></td>
+      <td>The style that is applied to each individual item.</td>
```
thevortexcloud, line 43: This is not technically correct. It can be used to apply to an item but really it will work with any child element of the items control

docs/reference/controls/itemscontrol.md
```diff
@@ -89,7 +115,61 @@ public class Crockery
 
 The view resizes horizontally, but content is hidden when it is too high. This control does not have a built-in scrollbar (unlike `ListBox`).
 
-<img src={ItemsControlScreenshot} alt="" />
+<img src={example01Screenshot} alt="" />
+
+## Example 02
+This example bind an observable collection of tile items to an items control. The ItemPanel is set to a canvas and we use a style to position the Tile within the canvas.
```
thevortexcloud, line 121: There is a typo here, it should be "binds" not "bind". I would also avoid saying tiles and instead say `Rectangle`s since it makes it much clearer what the code is doing.

docs/reference/controls/itemscontrol.md
```diff
@@ -89,7 +115,61 @@ public class Crockery
...
+
+namespace AvaloniaControls.ViewModels
+{
+    public class MainWindowViewModel : ViewModelBase
+    {
+        public ObservableCollection<Tile> TileList { get; set; }
+        
+        public MainWindowViewModel()
+        {
+            TileList = new ObservableCollection<Tile>(new List<Tile>
```
thevortexcloud, line 157: I am not sure if this is a good idea putting this VM in the examples. It sort of _loosely_ breaks MVVM by making the VM directly care about presentation. But other people may say otherwise.

docs/reference/controls/itemscontrol.md
thevortexcloud: I would reword it entirely to something like "The item template, contains a DataTemplate which will be applied to individual items and can be used to change how items look". That way it better explains why someone would even use it and what it does.

docs/reference/controls/itemscontrol.md
cerobe: I just did the same things as the already existing example in the page. Happy to change that. What would be better? (I'm happy to change the existing example too for consistency)

timunie: I think in that case it is a valid option. Could be a representation of whatever. A pin mark on a map or similar. So I think this is okay, but probably some comments are a good idea.

cerobe: all right, what comment would be the best to add?

cerobe:
Updated the PR according to the feedback. Waiting for the alignment on the last review, should we put the VM in the example and put a comment about this loosely breaking MVVM.

docs/reference/controls/itemscontrol.md
```diff
@@ -118,7 +118,7 @@ The view resizes horizontally, but content is hidden when it is too high. This c
 <img src={example01Screenshot} alt="" />
 
 ## Example 02
-This example bind an observable collection of tile items to an items control. The ItemPanel is set to a canvas and we use a style to position the Tile within the canvas.
+This example binds an observable collection of Rectangles (which are bound to a tile VM) to an ItemsControl. The ItemPanel is set to a canvas and we use a style to position the Tile within the canvas.
```
thevortexcloud, line 121: You are not positioning the tile, you are positioning the rectangle based on the title data. It's an important semantic distinction to make for this sort of thing and making that clear can help people who are not familiar with MVVM.

docs/reference/controls/itemscontrol.md
```diff
@@ -32,15 +32,15 @@ You will probably use these properties most often:
     </tr>
     <tr>
       <td><code>ItemsControl.ItemTemplate</code></td>
-      <td>Attached property element to contain the data template for an individual item.</td>
+      <td>The item template, contains a DataTemplate which will be applied to individual items and can be used to change how items look.</td>
     </tr>
     <tr>
       <td><code>ItemsControl.ItemPanel</code></td>
-      <td>Parent panel to display the items in. By default, this is a StackPanel</td>
+      <td>The container panel to place items in.</td>
```
thevortexcloud, line 39: It's probably still worth mentioning that the default is a stack panel.

docs/reference/controls/itemscontrol.md
thevortexcloud, line 121: It's a minor thing but can you change all the control references to use inline code blocks? It makes it clearer you are referring to an actual type. EG `ItemsControl` instead of ItemsControl.

thevortexcloud (changes requested):

thevortexcloud:
Apart from the things I mentioned and my previous concerns about the VM, it looks good.

cerobe:
Thanks for your detailed and high quality review @thevortexcloud. I've updated the PR with your latest feedback and closed them. Only the VM is still open and waiting for @timunie input.

thevortexcloud:
You are welcome! I am fan of good documentation. Which is why I try to make sure documentation is reasonably detailed and have some sort of logic behind what it's saying.

docs/reference/controls/itemscontrol.md
timunie: for example: `// This is just some sample data storing the coordinates directly inside the item-model. In your real-world App you may want to use a converter or similar to calculate the position`

docs/reference/controls/itemscontrol.md
```diff
@@ -89,7 +115,61 @@ public class Crockery
...
+```
+
+```csharp title='C# View Model'
+using AvaloniaControls.Models;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace AvaloniaControls.ViewModels
+{
+    public class MainWindowViewModel : ViewModelBase
```
thevortexcloud, line 151: I just noticed you have `ViewModelBase` here. If you don't need it, you should remove it. People are not guaranteed to have that.

docs/concepts/custom-itemspanel.md
```diff
@@ -0,0 +1,71 @@
+---
+description: CONCEPTS
+---
+
+import exampleScreenshot from '/img/concepts/customitemspanel/ItemsControl-with-ItemPanel-and-Styles.png';
+
+# Custom ItemsPanel
+
+the following built-in controls, display their items in a StackPanel by default:
```
thevortexcloud, line 9: This is not true for all of these panels. For example Carousel uses it's own special custom panel

https://github.com/thevortexcloud/Avalonia/blob/2a7b5568b7f61fe6df9c46a05bf1c9d474c9290e/src/Avalonia.Controls/Carousel.cs#L18-L23

docs/concepts/custom-itemspanel.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Custom ItemsPanel
+
+the following built-in controls, display their items in a StackPanel by default: 
+- [`ItemsControl`](../reference/controls/itemscontrol.md)
+- [`TreeView`](../reference/controls/detailed-reference/treeview-1.md)
+- [`Carousel`](../reference/controls/carousel.md)
+- [`Menu`](../reference/controls/menu.md)
+- [`ComboBox`](../reference/controls/combobox.md)
+- `ListBox`
```
thevortexcloud, line 15: Is there a reason ListBox specifically does not have a link?

docs/concepts/custom-itemspanel.md
```diff
@@ -0,0 +1,71 @@
...
+
+the following built-in controls, display their items in a StackPanel by default: 
+- [`ItemsControl`](../reference/controls/itemscontrol.md)
+- [`TreeView`](../reference/controls/detailed-reference/treeview-1.md)
+- [`Carousel`](../reference/controls/carousel.md)
+- [`Menu`](../reference/controls/menu.md)
+- [`ComboBox`](../reference/controls/combobox.md)
+- `ListBox` 
+
+You can customise this default behaviour via the property `ItemsPanel` of the control. The example below demonstrate a customisation of an `ItemsControls`
```
thevortexcloud, line 17: I would suggest moving this into some sort of introduction section and rewording it to something like:

> "All `ItemsControl`s have an item container panel which is used to layout their items. It is possible to override the type of panel used by the control to achieve custom/alternative layouts of items in a control. This document provides some examples showcasing how and why you would do this.

docs/reference/controls/carousel.md
```diff
@@ -10,6 +10,37 @@ The carousel has an items collection, and displays each item as a page, in seque
...
+<table>
+  <thead>
+    <tr>
+      <th width="316">Property</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>ItemsSource</code></td>
```
thevortexcloud, line 26: It might be better off putting this on the `ItemsControl` page and then putting a link to it here instead. These are shared properties with all items controls and it will save someone the trouble of trying to find all references to them if they ever need to updated. But I am also not too sure about this either.

thevortexcloud (changes requested):

docs/concepts/custom-itemspanel.md
cerobe: I could not find the page, happy to link to the right one, if you point it to me

thevortexcloud: https://docs.avaloniaui.net/docs/reference/controls/listbox

docs/concepts/custom-itemspanel.md
cerobe: Indeed, it's much better. You are clearly better with words than me.

docs/reference/controls/carousel.md
cerobe: There was no "useful properties" section in this page, so I've added the same as the ItemsControl for a bit of standardisation. I find useful for new joiner to have information directly in the page even at the cost of small duplication...but as always It's just a matter of preference and I'm happy to change it if needed

thevortexcloud: Yeah, I am really in two minds about this. The downside with a link is that someone might scroll past it and not find what they are looking for. Ideally you would be able to share the markdown between all the pages from a single source and then have the doc system merge it all together. Apparently it can be done with some plugin for Docusaurus. But I have no idea what's installed on the doc site currently and I can't say I have much experience configuring it either.

docs/concepts/custom-itemspanel.md
cerobe: ...I was blind, I'm not sure how I missed it. It's now updated and this on has also a VirtualizingStackPanel as default like the Carousel.

docs/reference/controls/carousel.md
timunie: This is out of scope here, but we are trying to figure out how we could maybe pull these docs from the controls code directly. For now I am happy with wither solution and I'd say let's keep it as it is now.

timunie (approved):
looks good to me. If later on we find anything to add or update, we can do another PR. 

Many thanks for your efforts and contribution 💐

docs/reference/controls/carousel.md
thevortexcloud: I imagine it's just a matter of enabling [this](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#generatedocumentationfile) and parsing the output somehow. But nice to know that is being considered!

thevortexcloud:
FYI, the NPM build is now failing seemingly as a result of some change here:

>[success] [webpackbar] Client: Compiled with some errors in 35.62s
Error:  Client bundle compiled with errors therefore further build is impossible.
Error: Docs markdown link couldn't be resolved: (../reference/controls/itemscontrol.md) in "/home/runner/work/avalonia-docs/avalonia-docs/docs/concepts/custom-itemspanel.md" for version current
    at Array.forEach (<anonymous>)

--------------------------

Module not found: Error: Can't resolve '/img/gitbook-import/assets/items.gif' in '/home/runner/work/avalonia-docs/avalonia-docs/i18n/ru/docusaurus-plugin-content-docs/current/reference/controls'
Error: Process completed with exit code 1.

thevortexcloud:
> Tht should be fine...I guess, I could not check as I changing the language when running locally is not working for me (am I doing something wrong by using the language listbox?)

Not sure, I have never run the docs site locally before. In any case, it's still failing with the same error.

```
 [success] [webpackbar] Client: Compiled with some errors in 35.17s
Error:  Client bundle compiled with errors therefore further build is impossible.
Error: Docs markdown link couldn't be resolved: (../reference/controls/itemscontrol.md) in "/home/runner/work/avalonia-docs/avalonia-docs/docs/concepts/custom-itemspanel.md" for version current
    at Array.forEach (<anonymous>)
```

cerobe:
hum, I'm at a loss here. The links seems correct and it's working locally (I know it's programmer meme). Any Idea what is wrong?

timunie:
@cerobe probably case sensitive. Other than that I need to take a look next week or ask the team.

cerobe:
Have you been able to look at the issue @timunie ?

maxkatz6:
Build is failing with
Error: Docs markdown link couldn't be resolved: (./../reference/controls/itemscontrol.md) in "/home/runner/work/avalonia-docs/avalonia-docs/docs/concepts/custom-itemspanel.md" for version current

timunie:
@cerobe looks like omitting the .md extension helps. https://github.com/AvaloniaUI/avalonia-docs/pull/341/commits/1636b6b41e909fd6f03ab96dd95edfc94851cbdc 

I HAVE NO IDEA WHY AND IF LINKS WORKS AFTERWARDS. But it is the only option I found so far. 🤷

thevortexcloud:
If you don't have the extension it might not get compiled based on what I saw in the docasaurus docs.

https://docusaurus.io/docs/markdown-features

Has anybody tested if an absolute link works? It's not ideal but it's better than the current situation.

cerobe:
@timunie , weird one, should I try to remove the .md extension on all links in this page then?

timunie:
> should I try to remove the .md extension on all links in this page then?

Yes, please let us give that a try.

thevortexcloud:
Good news, looks like the build worked!

cerobe:
That was completely unexpected, but nice that you found that. WE should be able to merge this PR soon...Finally.

timunie:
That was an interesting drive 👀 

Thank you for your contribution and patience 🙏

thevortexcloud:
It's live and it looks like the links work.

https://docs.avaloniaui.net/docs/concepts/custom-itemspanel

PR #339, opened 2024-02-03
Update how-to-implement-dependency-injection.md

FlatlinerDOA:
Wrote up a simple tutorial for using DI in Avalonia (with some AI help).

### DIFF
--- a/docs/guides/implementation-guides/how-to-implement-dependency-injection.md
+++ b/docs/guides/implementation-guides/how-to-implement-dependency-injection.md
@@ -13,3 +13,149 @@ This guide will show you how to use Dependency Injection (DI) with _Avalonia UI_
 
 <img src={MvvmArchitectureDiagram} alt=""/>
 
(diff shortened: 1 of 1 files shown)


thevortexcloud:
Can you please proof read the document? It has typos and seemingly like most AI generated content is fairly hard to follow. For example you for some reason create a type called `IMyService` but then register `GuitarTabsService` which is not mentioned anywhere else in the code and some of the comments for example mention things that are just never explained or mentioned again. 

On another note: Avalonia already uses Splat internally for it's DI. It's possible to override the Splat container and use something else, such as MS DI. At which point Avalonia itself will directly use your container for its services. You can also register things directly with the Splat container without changing anything else. It's probably worth mentioning that.

An example of also how you would configure the container to work in a cross platform way would also be nice. EG on desktop platforms and on mobile ones sharing the same core code base.

I am also not clear what Avalonia's stance on AI generated content is. There are some possible legal issues around if you even have the legal right to contribute AI generated content.

PR #345, opened 2024-02-09
fix: Example Style for DataValidationErrors

workgroupengineering:
When CompiledBinding is enabled XamlC throws `Unable to infer DataContext type for compiled bindings nested within this element.`

To avoid adding `x:DataType="{x:Type x:Object}"` to DataTemplate


Arhell (approved):

maxkatz6:
Hm, I think ErrorTemplate property should have `[DataTypeAttribute(typeof(Object))]` on it. Though not sure if XamlX will handle it properly yet, but should.

maxkatz6 (approved):

PR #349, opened 2024-02-12
Update add-content-to-dialog.md

haekenssam:
<Window ...
xmlns:views="using:Avalonia.MusicStore.Views" >

Doesn't work 
change to "using:Avalonia.MusicStore"

### DIFF
--- a/docs/tutorials/music-store-app/add-content-to-dialog.md
+++ b/docs/tutorials/music-store-app/add-content-to-dialog.md
@@ -59,7 +59,7 @@ To add the user control, follow this procedure:
 
 ```xml
 <Window ...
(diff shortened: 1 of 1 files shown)


timunie:
In which case doesn't it work? I used the official repo yesterday and had no such issue. Please add details about the issue you're facing.

maxkatz6 (approved):

Mike-Plouffe:
@maxkatz6 Why was this approved? I am following the tutorial now and I need the .views in order for the code to function. When I pull up the completed github code, it also still contains .views there - https://github.com/AvaloniaUI/MusicStoreTutorial/blob/master/Avalonia.MusicStore/Views/MusicStoreWindow.axaml

PR #359, opened 2024-02-17
Adding content for the Dependency Injection guide

cerobe:
Creating a new PR to update the guide https://docs.avaloniaui.net/docs/guides/implementation-guides/how-to-implement-dependency-injection

There is already a PR https://github.com/AvaloniaUI/avalonia-docs/pull/339 but it seems stale.

I tried to create the simplest guide I could to help the reader grasping the essential. 

Do not hesitate to feedback, I'm happy to make as much update as needed to have the best documentation possible.


docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
 
 # How To Implement Dependency Injection
```
thevortexcloud, line 9: Please add an introduction explaining what DI is and when you would use it. It would help greatly with people who have never heard of it before.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
 
-Content in preparation.
+:::info
+For views you can make a simple [ViewLocator](../../concepts/view-locator) which will automatically resolve the view from the Locator.
```
thevortexcloud, line 11: Please don't mention this unless you are going to show how to do this. It's somewhat extraneous information and will likely just frustrate/confuse people. Further more since most people are care about VM injection, this can probably go at the end. 

But adding some short sample code with a custom view locator should be fairly straight forwards.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+:::
 
 This guide will show you how to use Dependency Injection (DI) with _Avalonia UI_ and the MVVM pattern.
```
thevortexcloud, line 14: If you are writing a proper introduction (as I mentioned in another review issue) then this can be removed.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
 # How To Implement Dependency Injection
 
-Content in preparation.
+:::info
+For views you can make a simple [ViewLocator](../../concepts/view-locator) which will automatically resolve the view from the Locator.
+:::
 
 This guide will show you how to use Dependency Injection (DI) with _Avalonia UI_ and the MVVM pattern. 
 
 <img src={MvvmArchitectureDiagram} alt=""/>
```
thevortexcloud, line 16: Do you still need this picture? I don't think you have referenced anything in it anywhere.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
 
+Let's assume that your MainViewModel as a dependency on MyService. The contructor for ViewModel should looks like this:
```
thevortexcloud, line 18: You have multiple typos here. It should be:

"Let's assume that your MainViewModel **has** a dependency on MyService"
and
"The **constructor** for ViewModel should looks like this:"

There is also some awkward phrasing with "The constructor for ViewModel should looks like this:". It probably needs an "the" before "ViewModel".

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+```csharp
+private readonly MyService _myService;
```
thevortexcloud, line 20: Typically you would inject interfaces, not concrete types. You lose one of the main benefits of DI by doing that which is easy unit testing. This is technically fine, but may lead to people learning the wrong lesson.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+
+public MapsViewModel(MyService myService)
```
thevortexcloud, line 22: Is there a particular reason this is called `MapsViewModel`? It does not seem relevant to anything here. I would suggest using a more generic name if it's not needed.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+dotnet add package Microsoft.Extensions.DependencyInjection
+```
+
+## Step 2: ServiceCollectionExtensions 
+The following code is creating an extension for IServiceCollection. This is where you will register all your common dependencies. It's recomended to create that class in the shared project used by all avalonia target platform.
+
+```csharp
+public static class ServiceCollectionExtensions {
+    public static void AddCommonServices(this IServiceCollection collection) {
+        collection.AddTransient<MainViewModel>();
```
thevortexcloud, line 40: You have not defined `MainViewModel` instead you have declared `MapsViewModel`.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+Let's assume that your MainViewModel as a dependency on MyService. The contructor for ViewModel should looks like this:
+```csharp
+private readonly MyService _myService;
+
+public MapsViewModel(MyService myService)
+{
+    _myService = myService;
+}
+```
+These are the steps you need to do in order to resolve that dependencu using DI.
```
thevortexcloud, line 27: Typo in "dependencu`

thevortexcloud, line 27: Please add more information. The code you copied from me specifically is for using DI in a multi platform way across platform specific projects where you might have platform specific service implementations. There are simpler ways of doing this if you are only dealing with a single project/shared platform (such as desktops).

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+
+## Step 1: Install the NuGet package Microsoft.Extensions.DependencyInjection
+```shell
+dotnet add package Microsoft.Extensions.DependencyInjection
```
thevortexcloud, line 31: It's worth mentioning that there are many other DI providers available but the guide will be using Microsoft's DI. It's probably also worth mentioning that Avalonia also actually ships with the Splat DI container and it's possible to use it from user code. Although it can require additional setup to do so.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+```
+
+## Step 2: ServiceCollectionExtensions 
+The following code is creating an extension for IServiceCollection. This is where you will register all your common dependencies. It's recomended to create that class in the shared project used by all avalonia target platform.
```
thevortexcloud, line 35: Please explain _why_ someone would want to create this extension method. EG it should mention something about reducing code duplication between target platforms in a convenient way.

thevortexcloud, line 35: I would reword this from "common dependencies" to "platform independent shared dependencies" to make it clearer what it is actually doing. You can also say that the shared dependencies can themselves have dependencies on a platform specific dependency (which was the whole point of this example code in the first place). Although if you do that, please add an example of a platform specific service.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+public static class ServiceCollectionExtensions {
+    public static void AddCommonServices(this IServiceCollection collection) {
+        collection.AddTransient<MainViewModel>();
+        collection.AddTransient<MyService>();
+    }
+}
+```
+
+## Step 3: ServiceManager
+The following code create a ServiceManager class that is used to retrieve an instance of a given object.
```
thevortexcloud, line 47: There is some incorrect grammar here. It should be "The following code create**s**..."

thevortexcloud, line 47: This is not what the `ServiceManager` is doing. It's just a fairly simple way to create a singleton to manage a global DI container. The `Locator` is what you would use to resolve an object from the container. You also don't need this class at all if you are only dealing with a single platform. As the DI container can be set up completely during framework initialisation.

thevortexcloud, line 47: Please add a warning explaining why you should not use the locator unless you have no other option.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+```csharp
+public static class ServiceManager {
+    public static IServiceProvider Locator { get; private set; }
+
+    public static void InitialiseLocatorFromCollection(IServiceCollection collection) {
+        Locator = collection.BuildServiceProvider();
```
thevortexcloud, line 53: This code was fine for me to just quickly show off how to do this, but since people are probably going to copy and paste this a lot it's worth doing properly. As such I would suggest:

1. Add comments to the `Locator` property and the `InitialiseLocatorFromCollection`method
2. Make the `InitialiseLocatorFromCollection` method check if the locator has already been initialised
3. Add some functionality to dispose of the locator when the platform shutsdown as people may have dependencies that need to be properly cleaned up

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+```csharp
+class Program
+{
+    // Initialization code. Don't use any Avalonia, third-party APIs or any
+    // SynchronizationContext-reliant code before AppMain is called: things aren't initialized
+    // yet and stuff might break.
+    [STAThread]
+    public static void Main(string[] args)
+    {
+        var collection = new ServiceCollection();
```
thevortexcloud, line 70: Please read the follow up comments in the discussion you copied this all from, it was discussed not to create the container like this since it can cause issues.

https://github.com/AvaloniaUI/Avalonia/discussions/13455

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+        collection.AddCommonServices();
+
+        //Register platform specific services here with the collection
```
thevortexcloud, line 73: It would be better to show a more practical example of this instead of just a comment. I did the comment because I was somewhat lazy at the time.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+        => AppBuilder.Configure<App>()
+            .UsePlatformDetect()
+            .WithInterFont()
+            .LogToTrace();
+
+}
+```
+
+## Step 5: Modify App.axaml.cs
+We are using the ServiceManager class to retrieve the right viewModel.
```
thevortexcloud, line 92: Please use inline code blocks, EG `ServiceManager`

thevortexcloud, line 92: "viewModel` is capitalised weirdly.

thevortexcloud, line 92: Please add more detail explaining what you are trying to accomplish in this step.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -7,9 +7,120 @@ import MvvmArchitectureDiagram from '/img/gitbook-import/assets/image (3) (1) (2
...
+        AvaloniaXamlLoader.Load(this);
+    }
+
+    public override void OnFrameworkInitializationCompleted()
+    {
+        // Line below is needed to remove Avalonia data validation.
+        // Without this line you will get duplicate validations from both Avalonia and CT
+        BindingPlugins.DataValidators.RemoveAt(0);
+
+        var vm = ServiceManager.Locator.GetRequiredService<MainViewModel>();
```
thevortexcloud, line 107: Add comments explaining what this code is doing.

thevortexcloud (changes requested):

thevortexcloud:
@Irasi Sorry for the ping, but seeing as you have actually used my example code, do you have any additional suggestions?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: The link redirect to a documentation page with sample code already. 

I see your point with frustrating/confusing people, so I'll remove this part.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I've copy/paste the introduction from the other PR, as it's seems ok.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: The link it points to contains nothing about using a view locator with with DI. It's just pure reflection. Which is why I said that. This page is purely about using DI with Avalonia.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: Indeed. Do you have service in mind that would be fit for the role? I haven't written a multi-platform app in Avalonia yet.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: I would add an extra sentence somewhere, along the lines of "It accomplishes this by creating discrete services that are passed around/created as needed".

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: Hmm, of the top of my head a simple thing that does not really mean anything would be to create a service that returns a number depending on the platform. EG 1 on Linux, 2 on Windows, etc. 

If you want a more practical thing then you could make a service that reads file permissions. .NET has different methods for reading Windows file permissions compared to Unix ones.  If you try to use one on the on the wrong platform it will throw an exception.  A simple `CanRead()` method which returns a value indicating if the program can read a file should be enough.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: Thanks, it's indeed a good improvement.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: Ok, can you check if what I have added is enough?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -5,7 +5,7 @@ title: How To Implement Dependency Injection
 
 # How To Implement Dependency Injection in Avalonia
 
-[Dependency injection (DI)](https://en.wikipedia.org/wiki/Dependency_injection) allows developers to write cleaner, more modular, and testable code. It is especially useful when you are developing multi-platform applications that need specific service implementations for each platform.
+[Dependency injection (DI)](https://en.wikipedia.org/wiki/Dependency_injection) allows developers to write cleaner, more modular, and testable code.It accomplishes this by creating discrete services that are passed around/created as needed.q It is especially useful when you are developing multi-platform applications that need specific service implementations for each platform.
```
thevortexcloud, line 8: You are missing a space before "it accomplishes". You also have a errant "q" before "It is especially useful"

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -85,7 +85,7 @@ class Program
 ```
 
 ## Step 5: Modify App.axaml.cs
-We are using the `ServiceManager` class to retrieve the right view model.
+Modiying the `App` class to use `ServiceManager` to retrieve instances of the data context with all their dependencies resolved via DI.
```
thevortexcloud, line 88: This feels less like a tutorial and more like a command? You could instead phrase this as "Next, the `App.xaml.cs` class should be modified to use the DI container. This will allow the previously registered view model to be resolved via the dependency injection container. The fully realised view model can then be set to the data context of the main view. "

thevortexcloud, line 88: If you decide to keep the existing phrasing, please fix the typo "Modiying".

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -100,6 +100,8 @@ public partial class App : Application
         // Without this line you will get duplicate validations from both Avalonia and CT
         BindingPlugins.DataValidators.RemoveAt(0);
 
+        // `ServiceManager` retrieve an instance of `MainViewModel`
```
thevortexcloud, line 103: I assume this is a copy paste error, but is there a reason you have markdown code blocks inside C# comments?

thevortexcloud, line 103: I would reword this to be far clearer so someone who has never used DI before can see more clearly what it does. Something like "The service manager's locator instance is used to resolve an instance of MainViewModel, with all the constructor depedencies of MainViewModel such as MyService that was configured during the container creation."

But there may be a better way of phrasing all of that.

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: I have added additional comments.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I wonder if it's worth it to add all this in this small guide. At the moment it seems we have the core concept of DI without too much fluff around. 

As a compromise we could maybe extend that comment to add a bit more info?

cerobe:
@thevortexcloud I've done little improvements: 
- to standardise the titles : starting with a verb for the action to do in that step.
- moving the sentence about the nugget package from the intro into to the corresponding step
- specified the location of where to add the new file in the context of a multi-platform app (please check I'm correct on that)

Question: I wonder it the multi-platform aspect is noticeable enough for the reader?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,127 @@ id: how-to-implement-dependency-injection
...
+private readonly MyService _myService;
+
+public MapsViewModel(MyService myService)
+{
+    _myService = myService;
+}
+```
+These are the steps you need to do in order to resolve that dependencu using DI.
+
+## Step 1: Install the NuGet package for DI
```
maxkatz6, line 24: Our first step and option should be on using DI without any IoC.
I.e. simple:
```csharp
var window = new MainWindow
{
    DataContext = new MapsViewModel(new MyService())
}
```
Or something like that. It also will serve a good example **why** people prefer IoC containers, to reduce boilerplate. 
All of the IoC libraries will get quickly opinionated.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
maxkatz6: Yes, Main method is not a universal entry point of the Avalonia app. 
It won't work with Designer (its entry point is BuildAvaloniaApp method). It also won't work with mobile - each platform has their own, like MainActivity on Android. 

Universal place would be Application class, which is always created. Coincidentally, it's a convention place to create a MainWindow, which is usually also a root where IoC is resolved.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
maxkatz6: > Avalonia also actually ships with the Splat DI container

It's not. Splat is a dependency of ReactiveUI, which is an optional choice for developers.
And Splat DI was created to serve ReactiveUI extensibility purposes, and in general is pretty poor IoC implementation. It probably shouldn't be mentioned at all.

maxkatz6: After MS.DI a popular second choice would be lightweight DryIoC or some older but heavier options like Autofac.
Possibly some newer source-generators based solutions could be a good choice for developers, who know what they are doing. Something like Pure.DI even has an avalonia sample - https://github.com/DevTeam/Pure.DI.

maxkatz6: I think it would be a fair game to provide a short example with MS.DI, and reference links to more options and detailed docs on MS.DI as well.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
Irasil: If I can share my thoughts, I'd like to see an example of how to set up services for different platforms. That's the main reason I picked Avalonia UI. Maybe you could make two examples...

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
Irasil: That would be very helpful, as I still have it in the main, since it works and I currently have other priorities.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: Huh, I guess I was mistaken. Good to know. But yes more general information about DI as well as links to more advanced usage of MS Di would also be good.

thevortexcloud:
> Question: I wonder it the multi-platform aspect is noticeable enough for the reader?

Considering I did not write any platform specific services when I wrote the code, I would say not. So far that code will work on all desktop platforms without any difference between them. It really does need an example of how you would change what services get registered depending on the platform.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
timunie: > [!TIP]
> This can also go into Avalonia.Samples and cross linked to the docs.

cerobe: Yes I think adding a more specific example for multi-platform DI in Samples would be better.

Here it's just a very simple guide for DI in Avalonia, which would have help me greatly a few weeks ago when I started with my desktop app. :D

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: Agree, I kept the example very simple to focus on the DI concept. I could add this in the step one but I would need to pull the whole app class to keep the context...which may be a bit overkill don't you think?

From my understanding, the idea here in this guide is to explain how install the nugget, what code to add and what to configure to have a skeleton of DI working. Am I correct?

cerobe: After a bit more thought I've added your suggestion to the step 1. Could you check If I have captured and articulated your point correctly please?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -14,11 +14,21 @@ Let's assume that your MainViewModel has a dependency on MyService. The construc
...
 private readonly MyService _myService;
 
-public MapsViewModel(MyService myService)
+public MainViewModel(MyService myService)
 {
     _myService = myService;
 }
 ```
+
+You could solve that by instanciating `MyService` and passing as a parameter everytime you need a `MainViewModel` instance, like this:
```
thevortexcloud, line 23: Typo in "instanciating", should be "instantiating".

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -14,11 +14,21 @@ Let's assume that your MainViewModel has a dependency on MyService. The construc
...
+```csharp
+var window = new MainWindow
+{
+    DataContext = new MainViewModel(new MyService())
+}
+```
+but this will get quickly difficult as when your application scale and your classes have more dependencies, hence the need of DI.
```
thevortexcloud, line 30: Capitalise "b"

thevortexcloud, line 30: Also for some reason Github seems to think this is C#, you may not have closed a code block properly? I can't see anything obvious though.

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I've jump line before and after the code block like I've done for the ones in the page. Hopefully it should fix it

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
maxkatz6: I don't see changes in the step 1?

> From my understanding, the idea here in this guide is to explain how install the nugget

I would say, the idea should be to show options that users have to setup DI, referencing some links. And MS.DI as a practical example with simple steps.

maxkatz6: Also see https://github.com/AvaloniaUI/avalonia-docs/pull/359#discussion_r1493882114
And https://github.com/AvaloniaUI/avalonia-docs/pull/359#discussion_r1493880450

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: It was from my code, and I missed to change this, it's now corrected

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I was not aware that there is a simplest way for this scenario to be honest. The app I've created using the wizard seems to be setup like a multi-platform one even if it's only desktop. I'm happy to add more info but you'll need to give me more details as I lack experience on this.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I've added mention to the DI you listed @maxkatz6

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I've tried to modify the example according the discussion and use AfterSetup(). It's working on my small app but can you check this is correct? Should I also add a comment explaning what you just said above @maxkatz6 ?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I've just done 2. but will need your help for 1. and 3.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,144 @@ id: how-to-implement-dependency-injection
...
+
+## Step 3: Add ServiceManager
+The following code creates a ServiceManager class that is used to retrieve an instance of a given object. It's recomended to create that class in the shared project used by all avalonia target platform.
+
+```csharp
+public static class ServiceManager {
+    public static IServiceProvider Locator { get; private set; }
+
+    public static void InitialiseLocatorFromCollection(IServiceCollection collection) {
+        if (Locator != null)
```
thevortexcloud, line 64: Isn't this backwards? Surely you want to set it if the locator is null.

thevortexcloud, line 64: I would expect this to throw an exception instead of just silently failing if the locator can't be set.

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: indeed it's working on my side, but it should not...

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -20,14 +20,16 @@ public MainViewModel(MyService myService)
 }
 ```
 
-You could solve that by instanciating `MyService` and passing as a parameter everytime you need a `MainViewModel` instance, like this:
+You could solve that by instantiating `MyService` and passing as a parameter everytime you need a `MainViewModel` instance, like this:
```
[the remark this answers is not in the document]
cerobe: It's long as you said but as it will be helpful as it's explicitly listing the problem instead of just alluding to them. For me both are correct it's just depends of the tone of voice and the target audience of the documentation.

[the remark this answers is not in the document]
cerobe: I've been able to make it easier to read by using list. I think it's a nice addition, thanks for the feedback

cerobe:
@thevortexcloud I kinda understand it, but if I do this in the example, I should probably remove the AfterSetup() I've added too. Am I correct?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: It wont throw an exception here since you are ignoring it when the condition is false.

thevortexcloud:
That's only if you are not doing the locator stuff to make the container configurable from all projects. The problem with sticking it all into the app class is that it's in the shared project. Which means it can't access any of the platform specific APIs provided by the platform specific TFMs. Which means you can't write services that use any of those APIs to put into the container. As such the workaround is to instead construct the container one level higher, in the platform specific projects and then resolve the needed services in the shared project. Which is exactly what my code does.

Also because Microsoft decided to make things more painful it also can't directly depend on anything that is in a platform specific in a project with a generic TFM project (there is no real reason for it from what I can tell, you can easily write the exact same functionality yourself in many cases which will compile fine). You can however indirectly depend on it by passing in types created in a platform specific TFM to a generic TFM, even if they use a platform specific API. As a result there is a fairly specific hierarchy you have to use for your project structure, even if it makes your code significantly more complicated as a result. None of that matters though if you are only using a single TFM.

Showing both approaches to setting up DI is fairly useful as far as I am concerned. It's also a pretty good demonstration of how to work around .NET limitations.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
thevortexcloud: 1.  ```csharp
    /// <summary>
    /// Handles global dependency injection related functionality
    /// </summary>
    public static class ServiceManager {
    /// <summary>
    /// A singleton instance of the service provider
    /// </summary>
    /// <remarks>
    /// This can be used to directly resolve services from the container. However doing so should be avoided whenever
    /// possible to avoid creating a dependency on the locator instance itself. In the vast majority of cases, the service
    /// container will be able automatically resolve services for you 
    /// </remarks>
    public static IServiceProvider? Locator { get; private set; }

    /// <summary>
    /// Initialise <see cref="Locator"/> by building the <paramref name="collection"/> into a <see cref="IServiceProvider"/>
    /// </summary>
    /// <param name="collection">The collection to create a service provider from</param>
    /// <remarks>This method may only ever be called ONCE</remarks>
    public static void InitialiseLocatorFromCollection(IServiceCollection collection) {
        //Check if the locator has been initialised previously
        if (Locator is null) {
            //Build the locator from the given collection
            Locator = collection.BuildServiceProvider();
            return;
        }

        //The locator has already been set, throw an error since this may lead to an inconsistent application state if
        //dependencies have already been resolved from the old locator
        throw new ApplicationException("The locator cannot be set more than once without explicitly disposing it!");
    }
    }

    /// <summary>
    /// Disposes the <see cref="Locator"/> instance and sets it back to a state where it can be reinitialised
    /// </summary>
    /// <remarks>This method will typically only ever be called on an application exit</remarks>
    public static void DisposeLocator() {
        if (Locator is not null && Locator is IDisposable disposable) {
            disposable.Dispose();
            Locator = null;
        }
    }
    ```
2. Put this somewhere that will be executed during boot where you have access to the Application lifetime (EG app class). It wont always work since you can't run code if something kills your app, but it should cover most graceful exits for lifetimes that support it. You may need to slightly tweak it depending on where you put it.
   ```csharp
       if (builder.Instance!.ApplicationLifetime is IControlledApplicationLifetime controlledApplicationLifetime) {
                    controlledApplicationLifetime.Exit += (sender, args) => {
                        ServiceManager.DisposeLocator();
                    };
                } 
   ```

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,149 @@ id: how-to-implement-dependency-injection
...
+```csharp
+private readonly IMyService _myService;
+
+public MainViewModel(IMyService myService)
+{
+    _myService = myService;
+}
+```
+
+Typically you would directly instantiate `MyService` and pass it into `MainViewModel`, like this:
```
thevortexcloud, line 23: You may need to do a very minimal implementation of the service to show `MyService` and prevent people complaining about the example not compiling. Otherwise you will need to reword everything else to refer to the interface.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,149 @@ id: how-to-implement-dependency-injection
...
+var window = new MainWindow
+{
+    // Note: we assume that MyService implements IMyService
+    DataContext = new MainViewModel(new MyService())
+}
+```
+
+ This works great for simple constructors that are not used very often and don't change. But this technique does not scale very well because:
+- The more dependencies your constructor has the more things you will need to instantiate and pass in. Instantiating the dependencies locally (such as by doing new MainViewModel(new MyService())) results in direct rigid coupling to a specific instance of the dependencies.
+- Similarly if MainViewModel creates it's dependencies itself (such as in the constructor body) it also becomes directly coupled to the creation of the dependencies which can result in mostly the same problems.
```
thevortexcloud, line 35: II probably should have proof read this more before I shared it. But it should be:

"its" not "it's". It also probably needs a comma after 
> (such as in the constructor body)

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,149 @@ id: how-to-implement-dependency-injection
...
+{
+    // Note: we assume that MyService implements IMyService
+    DataContext = new MainViewModel(new MyService())
+}
+```
+
+ This works great for simple constructors that are not used very often and don't change. But this technique does not scale very well because:
+- The more dependencies your constructor has the more things you will need to instantiate and pass in. Instantiating the dependencies locally (such as by doing new MainViewModel(new MyService())) results in direct rigid coupling to a specific instance of the dependencies.
+- Similarly if MainViewModel creates it's dependencies itself (such as in the constructor body) it also becomes directly coupled to the creation of the dependencies which can result in mostly the same problems. 
+- Furthermore, if the object is instantiated in many places, every single reference to any of the dependencies would also need to be updated should the dependencies of `MainViewModel` ever change (such as by requiring additional dependencies) or require a different implementation of a dependency.
```
thevortexcloud, line 36: > (such as by requiring additional dependencies)

Move the closing `)` to the end of the sentence.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,149 @@ id: how-to-implement-dependency-injection
...
+- The more dependencies your constructor has the more things you will need to instantiate and pass in. Instantiating the dependencies locally (such as by doing new MainViewModel(new MyService())) results in direct rigid coupling to a specific instance of the dependencies.
+- Similarly if MainViewModel creates it's dependencies itself (such as in the constructor body) it also becomes directly coupled to the creation of the dependencies which can result in mostly the same problems. 
+- Furthermore, if the object is instantiated in many places, every single reference to any of the dependencies would also need to be updated should the dependencies of `MainViewModel` ever change (such as by requiring additional dependencies) or require a different implementation of a dependency. 
+
+Dependency injection solves these problem by abstracting away the creation of objects and their dependencies. This allows for well encapsulated services to be used and registered that will be automatically passed into any other service that is registered to use them.
```
thevortexcloud, line 38: I would probably remove

>and registered

It makes the sentence flow much better.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,149 @@ id: how-to-implement-dependency-injection
...
+
+Dependency injection solves these problem by abstracting away the creation of objects and their dependencies. This allows for well encapsulated services to be used and registered that will be automatically passed into any other service that is registered to use them. 
+
+See below the steps you need to do in order to resolve that dependencu using DI.
+
+## Step 1: Install the NuGet package for DI
+There are many dependency injection (DI) container providers available ([DryIoC](https://github.com/dadhi/DryIoc), [Autofac](https://github.com/autofac/Autofac), [Pure.DI](https://github.com/DevTeam/Pure.DI)) but this guide will only focus on `Microsoft.Extensions.DependencyInjection`. This is a lightweight, extensible dependency injection container that is part of the .NET Framework. It provides an easy-to-use and convention-based way to add DI to .NET applications, including Avalonia-based desktop applications.
```
thevortexcloud, line 43: Slight correction. Microsoft DI is not part of the framework, which is why it's in an optional extension library. Also saying .NET Framework is misleading. .NET Framework specifically refers to the Windows only implementation of .NET, not the modern version of it which runs on more than just Windows.

thevortexcloud (changes requested):

cerobe:
I agree that the showing both approached to setup a DI will be useful. 
I will rework this guide to focus on a single TFM, clean the code sample and link to another page for multiple TFM (this page will probably under construction).

cerobe:
@thevortexcloud , I've updated the guide to focus on on the simpler approach. Let me know what you think about it

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,142 @@ id: how-to-implement-dependency-injection
...
+    {
+        // Line below is needed to remove Avalonia data validation.
+        // Without this line you will get duplicate validations from both Avalonia and CT
+        BindingPlugins.DataValidators.RemoveAt(0);
+
+        // Register all the services needed for the application to run
+        // If in design mode we register different set of services using the extension method from step 2
+        var collection = new ServiceCollection();
+        collection.AddCommonServices();
+        if (Design.IsDesignMode)
```
thevortexcloud, line 113: I should mention: Since you decided to go with the whole design/real services from my example, you can't use constructor injection with the previewer, at least out of the box. It wont work in most cases. You need to use a markup extension to resolve the services and set the design time data context. I excluded it from my example to avoid overcomplicating it since it also brings back in the locator as there is no way to pass a reference to the container into the markup extension that I know of. I was fairly lazy here so I just shoved the locator into the App class. But that service manager class would also work pretty well here.

```csharp
    public class DesignLocator : MarkupExtension {
        #region Public properties
        public Type Type { get; init; }
        #endregion

        #region Public methods
        public override object ProvideValue(IServiceProvider serviceProvider) {
            if (!Design.IsDesignMode) {
                throw new InvalidOperationException();
            }

            return App.Locator.GetRequiredService(this.Type);
        }
        #endregion
    }
```

You use it like this:

```xaml
        Design.DataContext="{markupExtensions:DesignLocator Type={x:Type vm:MainWindowViewModel}}"
```

See https://github.com/AvaloniaUI/Avalonia/discussions/13743 for the source for this idea.

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,142 @@ id: how-to-implement-dependency-injection
...
+- The more dependencies your constructor has the more things you will need to instantiate and pass in. Instantiating the dependencies locally (such as by doing new MainViewModel(new MyService())) results in direct rigid coupling to a specific instance of the dependencies.
+- Similarly if MainViewModel creates its dependencies itself, (such as in the constructor body) it also becomes directly coupled to the creation of the dependencies which can result in mostly the same problems. 
+- Furthermore, if the object is instantiated in many places, every single reference to any of the dependencies would also need to be updated should the dependencies of `MainViewModel` ever change (such as by requiring additional dependencies or require a different implementation of a dependency). 
+
+Dependency injection solves these problem by abstracting away the creation of objects and their dependencies. This allows for well encapsulated services to be used that will be automatically passed into any other service that is registered to use them. 
+
+See below the steps you need to do in order to resolve that dependency using DI.
```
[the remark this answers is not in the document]
cerobe: I think it flows with the previous one which is listed the problems. I'm happy to removed it if you think it's needed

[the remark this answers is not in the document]
thevortexcloud: The thing is you have already implied the steps by giving the step number. It's kind of like a chapter in a book directly saying the next chapter is about to start. It would be better to put this closer to the introduction if you wanted to keep it.

thevortexcloud (changes requested):

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
cerobe: I'm a bit uncertain on how to test this on my local sample, so I prefer to remove the design part. It's probably another guide to create in the documentation.

cerobe:
@thevortexcloud sorry for the long answer, I had a lot IRL stuff to do. I have now worked on the remaining comment, I think this PR is ready to be merge but can you have another look please?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,123 @@ id: how-to-implement-dependency-injection
...
+- Furthermore, if the object is instantiated in many places, every single reference to any of the dependencies would also need to be updated should the dependencies of `MainViewModel` ever change (such as by requiring additional dependencies or require a different implementation of a dependency). 
+
+Dependency injection solves these problem by abstracting away the creation of objects and their dependencies. This allows for well encapsulated services to be used that will be automatically passed into any other service that is registered to use them. 
+
+## Step 1: Install the NuGet package for DI
+There are many dependency injection (DI) container providers available ([DryIoC](https://github.com/dadhi/DryIoc), [Autofac](https://github.com/autofac/Autofac), [Pure.DI](https://github.com/DevTeam/Pure.DI)) but this guide will only focus on `Microsoft.Extensions.DependencyInjection` which is a lightweight, extensible dependency injection container. It provides an easy-to-use and convention-based way to add DI to .NET applications, including Avalonia-based desktop applications.
+
```
thevortexcloud, line 64: I think you need a simple sentence here along the lines of "run the following command in a terminal inside your project directory to install the DI package". Otherwise you are not really giving much information about what to do with the command.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,123 @@ id: how-to-implement-dependency-injection
...
+```shell
+dotnet add package Microsoft.Extensions.DependencyInjection
+```
+
+## Step 2: Add ServiceCollectionExtensions 
+The following code is creating an extension for IServiceCollection.
```
thevortexcloud, line 70: I would add change this to
> The following code is creating an extension method for `IServiceCollection` that will register services to our service collection and make them available for injection.

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,123 @@ id: how-to-implement-dependency-injection
...
+    public static void AddCommonServices(this IServiceCollection collection) {
+        collection.AddSingleton<IRepository, Repository>();
+        collection.AddTransient<BusinessService>();
+        collection.AddTransient<MainViewModel>();
+    }
+}
+```
+
+## Step 3: Modify App.axaml.cs
+Next, the App.xaml.cs class should be modified to use the DI container. This will allow the previously registered view model to be resolved via the dependency injection container. The fully realised view model can then be set to the data context of the main view.
```
thevortexcloud, line 83: Put "App.xaml.cs" into an inline code block.

thevortexcloud (changes requested):

thevortexcloud:
> sorry for the long answer, I had a lot IRL stuff to do

All good. It looks pretty good to me except for some minor stuff I mentioned.

cerobe:
Thanks @thevortexcloud .
@timunie can you have a look too and  approve for a merge into master?

docs/guides/implementation-guides/how-to-implement-dependency-injection.md
```diff
@@ -3,13 +3,125 @@ id: how-to-implement-dependency-injection
...
+    public override void Initialize()
+    {
+        AvaloniaXamlLoader.Load(this);
+    }
+
+    public override void OnFrameworkInitializationCompleted()
+    {
+        // Line below is needed to remove Avalonia data validation.
+        // Without this line you will get duplicate validations from both Avalonia and CT
+        BindingPlugins.DataValidators.RemoveAt(0);
```
maxkatz6, line 99: This 
```suggestion
        // If you use CommunityToolkit, line below is needed to remove Avalonia data validation.
        // Without this line you will get duplicate validations from both Avalonia and CT
        BindingPlugins.DataValidators.RemoveAt(0);
```

This part is unrelated to the documentation, so worth to make it a bit clearer why it's here.

maxkatz6 (approved):
LGTM, thank you!

timunie:
> @timunie can you have a look too and approve for a merge into master? 

Yes, can do it after last hint of Max is added. Thanks for all your time spent here. It's already very informative and will help many ppl.

PR #366, opened 2024-02-19
Update docusaurus.config.js config to show version as 11.0.x

Fxplorer:
This will align with the previous version naming of 0.10.x and will help reduce confusion to new users.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -100,7 +100,7 @@ const config = {
           lastVersion: 'current',
           versions: {
             current: {
-              label: "11.0.0",
+              label: "11.0.x",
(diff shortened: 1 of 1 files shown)


maxkatz6:
But what about user confusion, when they use 11.0.9 version, but can only find docs for 11.0.0? I know it's the same in the docs, as we don't have any new major features in patch releases, but that's a point.

Fxplorer:
This is exactly what happened to me this weekend. My version installed is 11.0.9 but the docs say 11.0.0 with no option for a version higher.

maxkatz6 (approved):

PR #378, opened 2024-02-28
Create how-to-bind-multiple-properties.md

### LINKED ISSUE #292: MultiBinding need a documentation with examples
As well as some other missed markup extensions (though, MultiBinding is not a markup extension strictly speaking)

stevemonaco:
Creates a brief description of `MultiBinding` and `IMultiValueConverter`. Provides an example of how to use and implement each.

Title is a bit rough, but in-line with the other scenarios in Guides -> Data Binding.

Closes #292


docs/guides/data-binding/how-to-bind-multiple-properties.md
```diff
@@ -0,0 +1,102 @@
...
+    [ObservableProperty] private byte _red;
+    [ObservableProperty] private byte _green;
+    [ObservableProperty] private byte _blue;
+    [ObservableProperty] private byte _alpha;
+}
+
+public partial class MainViewModel : ViewModelBase
+{
+    [ObservableProperty] private ColorViewModel _color;
+}
```
maxkatz6, line 47: It's a bit iffy for me to rely on SourceGenerated view models as a sample. Since this concept isn't as familiar for many folks. But it's a nitpick.

sidebars.js
```diff
@@ -288,6 +288,7 @@ const sidebars = {
             'guides/data-binding/how-to-bind-to-a-collection',
             'guides/data-binding/binding-from-code',
             'guides/data-binding/how-to-create-a-custom-data-binding-converter',
+            'guides/data-binding/how-to-bind-multiple-properties',
```
maxkatz6, line 291: Since you added this page, doe ru and zh pages still work?

docs/guides/data-binding/how-to-bind-multiple-properties.md
stevemonaco: I thought the same. But I'm not sure what the lesser evil is here: use sourcegen for brevity, use an MVVM framework with full properties (Mvvm Toolkit vs ReactiveUI are still different), or plop in an entire custom `INotifyPropertyChanged` implementation? I do need to demonstrate *something* for any example to make sense.

sidebars.js
stevemonaco: I overlooked this, incorrectly believing that sidebars had localized versions. There must be a built-in fallback, because the localized versions still work.

docs/guides/data-binding/how-to-bind-multiple-properties.md
maxkatz6: Specifically with readonly view model - it can be just readonly for simplicity. But it's not always the case.

stevemonaco: True, but readonly eliminates the point of using a `MultiBinding`. Otherwise, using a `Binding` with a `Converter` for the full object is the right approach because the properties won't notify anyways.

maxkatz6 (approved):

docs/guides/data-binding/how-to-bind-multiple-properties.md
stevemonaco: Actually, I can rework this with data coming from 4 `TextBox` or 4 `NumericUpDown` controls and create a visual. I don't think the VM choice is a big deal, but View-only has obvious merits in being self-contained (besides the converter). I'll amend when I have time.

timunie: I think a comment which should clarify it and link to source gen docs should be enough imo

Arhell (approved):

docs/guides/data-binding/how-to-bind-multiple-properties.md
stevemonaco: Updated example to View-only, added visual

maxkatz6:
By itself, interactive sample is definitely nicer!

PR #381, opened 2024-03-05
Bind to a Control with casting of parents DataContext

CodeDevAM:
Adds explanation how to cast the DataContext of a parents Control to access custom properties.

### DIFF
--- a/docs/guides/data-binding/binding-to-controls.md
+++ b/docs/guides/data-binding/binding-to-controls.md
@@ -86,7 +86,7 @@ If you need to include a XAML namespace in the ancestor type, you separate the n
 </local:MyControl>
 ```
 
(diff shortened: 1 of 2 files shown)


stevemonaco (commented):
Some typos: parents -> parent's. it's -> its. an compile time error -> a compile-time error.

stevemonaco (changes requested):
Sorry, I missed one thing. You should use "```xml" instead of markup. Looks good otherwise.

CodeDevAM:
> Sorry, I missed one thing. You should use "```xml" instead of markup. Looks good otherwise.

But the rest of the file is also using "```markup" for XAML examples. Should the other occurances altered as well?

stevemonaco:
> But the rest of the file is also using "```markup" for XAML examples. Should the other occurances altered as well?

After checking, it seems there's autodetection leads to the same highlighting anyways, so I guess it doesn't matter. Maintainers need to do a global pass of fixups in a separate issue.

stevemonaco (approved):

PR #382, opened 2024-03-06
Add FallBackValues to Binding page

Tamnac:
I don't really know how this works for non string content, so someone may want to add on.

Also not entirely sure if this should go on this page or somewhere under How to guides > Data Binding > ...

Currently there's a note mentioning in passing under one of those pages, which is insufficient and I was unable to find it without help from the community on telegram. I think this is valuable enough to warrant a proper section.

### DIFF
--- a/docs/basics/data/data-binding/data-binding-syntax.md
+++ b/docs/basics/data/data-binding/data-binding-syntax.md
@@ -138,7 +138,7 @@ An example of a custom converter can bind an image file. For guidance on how to 
 You may optionally define fallback values in the XAML in case the binding fails.
 
 ```xml
-<TextBlock Text="{Binding Name, FallBackValue=Bob}"/>
+<TextBlock Text="{Binding Name, FallbackValue=Bob}"/>
 ```
 
 


timunie:
For non-string fallback:

```xml
<UserControl.Resources>
   <system:Int32 x:Key="MyFallback">42</system:Int32>
</UserControl.Resources>

<NumericUpDown Value="{Binding MyValue, FallbackValue={StaticResource MyFallback}}" />
```

Please try it on your own first, I have written that out of my head. Not that I made a mistake here.

stevemonaco:
FYI, I'm currently working to refresh this particular doc to clean up language, add all Binding properties, etc, and will PR when ready. I have not yet written an example for `FallbackValue` because it's hard to demonstrate without subproperty bindings. Not sure what other scenarios `FallbackValue` is used while using compiled bindings.

stevemonaco:
> @stevemonaco even a basic example is better than nothing. If you'll include it in your rewrite, I'll close this pr

I will. I'm waiting for clarification on a different binding topic (`UpdateSourceTrigger`) in the same document before submitting. As of now, the section looks like the following (might receive some editing for repetition):

### FallbackValue

`FallbackValue` is used when the property binding cannot be made or when a converter returns `AvaloniaProperty.UnsetValue`.

A common use case is when a parent property in a subproperty binding is `null`. If `Student` is `null` below, 
the `FallbackValue` will be used:

```xml
<TextBlock Text="{Binding Student.Name, FallbackValue=Cannot find name}"/>
```

:::tip
`ReflectionBinding` can bind to arbitrary types without regard to compile-time safety. When the binding cannot be made, 
`FallbackValue` may be useful to substitute a value.
:::

PR #383, opened 2024-03-07
Updates Release 11.0.7

lukewire129:
Since we don't know if the PR will be accepted or not, we first use the 
11.0.7 Release history only


Arhell (commented):
lgtm

Issue #384, opened 2024-03-07
11.1 documentation requests


maxkatz6:
Note: all new APIs can already be submitted with a PR, but with `<MinVersion>` tag, see https://github.com/AvaloniaUI/avalonia-docs/pull/380

XAML:
- [ ] Warnings support, configurability

New services:
- [ ] ILauncher
- [ ] IInputPane
- [ ] Application.TryGetService<IActivatableLifetime>()

New/updated controls:
- [ ] HyperlinkButton
- [ ] MenuItem ToggleType docs (not a new control, but whole new API surface for menus)
- [ ] TextBox, NumericUpDown and AutoCompleteBox API changes (do we need TextSelectionHandle docs?)
- [ ] Border.BackgroundSizing
- [ ] Extended WindowNotificationManager - see https://github.com/AvaloniaUI/Avalonia/pull/12178

Updated platform options:
- [ ] MacOSProperties
- [ ] Win32Properties
- [ ] X11Properties
- [ ] Updated LinuxFramebufferPlatformOptions, iOSPlatformOptions and Win32PlatformOptions (missing docs to begin with)

Misc:
- [ ] New Binding features: UpdateSourceTrigger
- [ ] New XYFocus configurability
- [ ] Document localization of built-in controls, like TimePicker. Custom ResourceProvider implementation. See https://github.com/AvaloniaUI/Avalonia/pull/13773
- [ ] Dispatcher UnhandledException and UnhandledExceptionFilter - keep it low key, as not recomended feature, but copy some docs from WPF.
- [ ] FontFeature, AppBuilder.WithSystemFontSource (missing more font specific docs)
- [ ] DataValidationErrors - ErrorConverterProperty
- [ ] TopLevel.AutoSafeAreaPaddingProperty
- [ ] New ManagedFileDialogOptions
- [ ] New ISkiaSharpPlatformGraphicsApiLease (missing many related graphics APIs)

TV Support:
- [ ] Apple TV setup
- [ ] Tizen TV setup
- [ ] Android TV setup

stevemonaco:
Reserve the `XYFocus` + general focus docs for me. Almost finished with the informational pass, then need to decide where examples may be necessary.

qq908382818:
I want to install the extension AvaloniaVS.VS2022.vsix without an internet connection. Is there an offline installation package available? Where can I download it?

stevemonaco:
[...]

PR #392, opened 2024-03-12
Create focus.md

stevemonaco:
Creates general documentation for all focus-related features I'm aware of, including XYFocus.

I wasn't able to think of a compelling example of tab navigation and most other features are "basic", so I only added an example for XYFocus. If more are added, Focus may need to be split into several documents.

Tab focus navigation may be a bit suspect/underdeveloped as I don't understand the area well. Particularly this statement is hand-wavy:

```
When `TabIndex` is equal, the order of appearance within the Visual Tree is used.
```

Part of: #384


docs/concepts/input/focus.md
```diff
@@ -0,0 +1,149 @@
...
+## FocusManager
+
+The `FocusManager` provides global access to focus functionality, such as retrieving the currently focused element or
+clearing focus. For additional information, see the [FocusManager docs](../services/focus-manager).
+
+## Tab Focus Navigation
+
+Focus navigation by tab occurs when the user presses tab on their keyboard. `InputElement`s with their `IsTabStop` property 
+set to `true` will be available for tab focus navigation. The `TabIndex` specifies the priority with lower numeric values being 
+navigated to first. When `TabIndex` is equal, the order of appearance within the Visual Tree is used.
```
maxkatz6, line 60: Alternatively, something like "When `TabIndex` is equal, the order in which control is definition used". It's not entirely true, but should be good enough simplification.

docs/concepts/input/focus.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Focus Pseudoclasses
+
+These pseudoclasses are helpful in creating visual indicators when styling `Control`s that are `Focusable`.
+
+| Pseudoclass    | Description                                                                    |
+|:---------------|--------------------------------------------------------------------------------|
+| :focus         | The Control has focus.                                                         |
+| :focus-within  | The Control has focus or contains a descendant that has focus.                 |
+| :focus-visible | The Control has focus and received it from Tab or Directional focus navigation |
```
maxkatz6, line 49: To be more specific, `focus-visible` tells that control should use focus styles. Like, text box changes its appearance when focused. Additionally, if control overrides `:focus-visible`, it might need to set FocusAdorner to null:
https://github.com/AvaloniaUI/Avalonia/blob/0cd220f6c2ae409f5a27c261d09c2eb5836ad2fe/src/Avalonia.Themes.Fluent/Controls/TextBox.xaml#L107

As otherwise double focus visuals would be applied.

docs/concepts/input/focus.md
```diff
@@ -0,0 +1,149 @@
...
+| Cycle                  | Cycles through and wraps back within its own items        |
+| Contained              | Stops at the beginning/end item                           |
+| Once                   | Container and children receive focus only once as a group |
+| None                   | Items will not be focused by tab navigation               |
+| Local                  | `TabIndex` is considered for items in local subtree only  |
+
+## Directional Focus Navigation <MinVersion version="11.1" />
+
+Focus navigation through `XYFocus` is a 2D directional scheme enabling spatial navigation from the focused control 
+towards another control to a cardinal direction: left, right, up, or down. `XYFocus.NavigationModes` is `Disabled` by
```
maxkatz6, line 77: By default, it's Gamepad | Remote. Which means, it is enabled only for these types of controllers.

docs/concepts/input/focus.md
```diff
@@ -0,0 +1,149 @@
...
+|:--------------|:------------------------------------------|
+| Disabled      | Any key device XY navigation is disabled. |
+| Keyboard      | Keyboard arrow keys can be used.          |
+| Gamepad       | Gamepad controller DPad keys can be used. |
+| Remote        | Remote control can be used.               |
+| Enabled       | All devices can be used.                  |
+
+:::warning
+`KeyDeviceType.Gamepad` is not yet supported, so `XYFocus` navigation is currently limited to `Keyboard` (arrow keys)
+and `Remote` (tvOS most commonly).
```
maxkatz6, line 90: Gamepad is technically supported, as Android and Tizen already can send this kind of input.  We don't have a proper Gamepad APIs, but that's a different story, not related to XY focus or KeyDeviceType.

docs/concepts/input/focus.md
```diff
@@ -0,0 +1,149 @@
...
+modify the `Slider.Value` instead of causing navigation. Pressing Enter a second time will end the interaction.
+
+```xml
+<Window.Styles>
+    <Style Selector=":is(Control).directional-navigation">
+        <Setter Property="XYFocus.NavigationModes" Value="Enabled" />
+        <Setter Property="XYFocus.UpNavigationStrategy" Value="Projection" />
+        <Setter Property="XYFocus.DownNavigationStrategy" Value="Projection" />
+        <Setter Property="XYFocus.LeftNavigationStrategy" Value="Projection" />
+        <Setter Property="XYFocus.RightNavigationStrategy" Value="Projection" />
```
maxkatz6, line 128: These properties are inheritable, so it might make more sense to set it on top level.

docs/concepts/input/focus.md
stevemonaco: Feels a bit too simple because that only makes sense in the context of a single XAML document, not necessarily across documents or anything created by code.

maxkatz6 (approved):

stevemonaco:
Thanks for the review. Fixed the errors, suggestions, and cleaned up some new nits after another proofread pass.

PR #403, opened 2024-03-14
Remove 0.10.x localized docs

### LINKED ISSUE #401: Remove localized v0.10.x docs
AFAIK, the v0.10.x docs do not have any localization work and are simply English doc copies which leads to extra work for no value when doing global upgrades, such as #400.

I propose removing these docs, ultimately acknowledging that they won't be localized. The version selector in the localized docs will still fallback to the English v0.10.x as discovered in the #378 review.

I have tested this locally, simply by deleting `version-0.10.x` directory and `version-0.10.x.json` within `docusaurus-plugin-content-docs`.

I can test + PR this.

stevemonaco:
Implemented by deleting `version-0.10.x` directory and `version-0.10.x.json` within docusaurus-plugin-content-docs. Fallback to English docs is automatic.

Tested locally with npm build. Seems ok.

Closes #401


maxkatz6:
It has weird conflicts?

stevemonaco:
Let me close and redo. I think I was operating off of a version before #400

PR #425, opened 2024-03-24
Adding KeyBinding and MouseBinding

rtazaki:
- Since I created a new page, I modified sidebars.js.
- I didn't know how to preview the Tabs tag, so I implemented it by feel.
- I've never used docusaurus before because I'm not confident in multiple languages, but is it something that document creators must do?


stevemonaco (changes requested):

stevemonaco:
>     * I've never used docusaurus before because I'm not confident in multiple languages, but is it something that document creators must do?

For changes more extensive than modifying/adding words in an existing doc, probably. The CI runs checks for valid syntax/imports, checks md anchor links, image links, etc.

Setting up Docusaurus is easy, even if you don't do much web like me. Check the readme on the main page of this repo. Basically: Install Node.js, clone your forked docs repo, npm install, npx docusaurus start. This will start a local web server in dev mode and Docusaurus will hot reload pages as you modify (except the sidebar, IIRC, need to restart for that). You ideally build locally (npm run build --if-present) like the CI does before committing as dev mode will miss certain things like broken anchors.

rtazaki:
> > ```
> > * By the way, multiple languages, but is it something that document creators must do?
> >   (i.e. do I need to sync the i18n folder?)
> > ```
> 
> If the PR is accepted, then localizers will work on localizing the text afterwards. There aren't any localized images. At some point soon, I will probably sync the diffs manually for non-localizable parts (ie. code).
> 
> I would prefer to see some reasoning. eg. Why `MouseBinding` doesn't work. (Is this because the control itself handles it before it can bubble? I'm not familiar with this exact area, so I won't accept/reject). We're also trying to move docs away from ReactiveUI usage though we haven't settled on any particular "standard" VM structure.

1. Mouse binding exists in standard WPF but is not implemented in AvaloniaUI.
Example of a document I wrote: want to execute a command by double-clicking the mouse.
=>
[MouseBinding](https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.mousebinding?view=windowsdesktop-8.0)
[Avalonia.Input](https://reference.avaloniaui.net/api/Avalonia.Input/)
`<MouseBinding Gesture="LeftDoubleClick">`

2. The goal is to show as "standard" an implementation as possible, without depending on any particular framework.
But, resulting in an unrealistic amount of code. It is trade off.
Example: 
- Is it troublesome to mash fruit in a "mortar and pestle"? 
- Is it troublesome to write a note saying "any brand is fine" every time you use a food processor?
=>
`[Reactive]`, `.Execute(null);`, `ReactiveCommand`
Three methods are typically difficult to implement: with CommunityToolkit, with ReactiveUI, and without both.
and even if you could implement it, the documentation won't help developers understand the framework.
(Is there an option to avoid it by not writing about ViewModel?)

3. Rather than writing abstract documentation, I would like to leave a concrete and short example. I hope it provides inspiration for what's possible, even if it becomes a legacy someday.

MikeCodesDotNET (approved):

PR #426, opened 2024-03-26
Update running-on-raspbian-lite-via-drm.md

lehrbua:
Visual Studio Error on return builder.StartLinuxDrm(args); The call does not clearly distinguish between the two methods

### DIFF
--- a/docs/guides/platforms/rpi/running-on-raspbian-lite-via-drm.md
+++ b/docs/guides/platforms/rpi/running-on-raspbian-lite-via-drm.md
@@ -156,7 +156,9 @@ public static int Main(string[] args)
     if (args.Contains("--drm"))
     {
         SilenceConsole();
-        return builder.StartLinuxDrm(args: args, card: "/dev/dri/card1", scaling: 1.0);
+        // By default, Avalonia will try to detect output card automatically.
+        // But you can specify one, for example "/dev/dri/card1".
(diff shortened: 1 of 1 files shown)


docs/guides/platforms/rpi/running-on-raspbian-lite-via-drm.md
```diff
@@ -156,7 +156,7 @@ public static int Main(string[] args)
     if (args.Contains("--drm"))
     {
         SilenceConsole();
-        return builder.StartLinuxDrm(args);
+        return builder.StartLinuxDrm(args: args, card: "/dev/dri/card1", scaling: 1.0);
```
maxkatz6, line 159:
```suggestion
        // By default, Avalonia will try to detect output card automatically.
        // But you can specify one, for example "/dev/dri/card1".
        return builder.StartLinuxDrm(args: args, card: null, scaling: 1.0);
```

maxkatz6: It's important to mention, as otherwise people will complain when card1 is not available for some reason.
See https://github.com/AvaloniaUI/Avalonia/blob/0b72cdb512187c8a4eded47e864e1194111e621f/src/Linux/Avalonia.LinuxFramebuffer/Output/DrmBindings.cs#L144-L164

maxkatz6 (approved):

PR #428, opened 2024-03-29
Fix Microsoft.Toolkit mentions

stevemonaco:
3 mentions of Microsoft.Toolkit were fixed to CommunityToolkit to reflect the new package name.

No localizer work required as I updated those docs, too.

### DIFF
--- a/docs/guides/data-binding/inotifypropertychanged.md
+++ b/docs/guides/data-binding/inotifypropertychanged.md
@@ -52,7 +52,7 @@ While implementing `INotifyPropertyChanged` isn't particularly complex, it can b
 Here's how you can achieve the same result as before, but using `ObservableObject`:
 
 ```csharp
-using Microsoft.Toolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.ComponentModel;
 
 public partial class MyViewModel : ObservableObject
 {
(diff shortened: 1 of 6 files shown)


docs/guides/data-binding/inotifypropertychanged.md
```diff
@@ -1,21 +1,29 @@
...
 id: inotifypropertychanged
-title: How to use INotifyPropertyChanged
+title: How to Use INotifyPropertyChanged
 ---
 
 ## Introduction
-The `INotifyPropertyChanged` interface is a critical component in the Model-View-ViewModel (MVVM) design pattern that helps create scalable and maintainable applications. By notifying that a property has been changed, it allows the View to update automatically, improving the communication between the components of your application.
+The `INotifyPropertyChanged` interface is a critical component in the Model-View-ViewModel (MVVM) design pattern that 
+helps create scalable and maintainable applications. By notifying that a property has been changed, it allows the View 
+to update automatically, improving the communication between the components of your application.
```
maxkatz6, line 9: Not sure, are these line breaks necessary here?

stevemonaco: No. They're only for text editor purposes and aren't visually represented in the site.

I should probably be using Soft Wrap in Rider, but there's been a mix of incredibly long lines and manually formatted line breaks in docs since before I started committing. There's no style guide here.

Should I revert that part or just use soft wrap in the future?

maxkatz6: Probably better to revert, keeping git diff to what was actually changed.

maxkatz6 (approved):

PR #434, opened 2024-04-01
Update navigate-views.md

bearyung:
the DataContext need to cast to MainWindowViewModel with ((vm:MainWindowViewModel)DataContext) to make use of AddItem method of it


stevemonaco:
I think this docs needs some additional written details regarding the syntax because this is an introductory tutorial with no prior mention of compiled bindings in its current version, AFAIK. It might be ok as a bandaid fix because ToDoList is being rewritten.

PR #430, opened 2024-03-30
docs: fix add-content-to-dialog.md

draptik:
Fix namespace

PR #440, opened 2024-04-03
Update add-content-to-dialog.md

bearyung:
Incorrect views path, should be 
using:Avalonia.MusicStore.Views
instead of 
using:Avalonia.MusicStore


stevemonaco:
This has an existing PR at #430 with some requested changes. I'll accept/merge whichever does it first and close the other.

PR #441, opened 2024-04-04
Update how-to-use-design-time-data.md

dbchandra23:
The sample code missed a few critical items to make it work properly in the design view, such as the Command properties. 

I have also added the xmlns and x:DataType to make the code works, given that Avalonia generated app now have CompiledBinding turned on by default.

I have also added the MainWindow and AppointmentView.axaml.cs changes to make the whole example compile from a base.


docs/guides/implementation-guides/how-to-use-design-time-data.md
```diff
@@ -47,16 +61,22 @@ public class DesignAppointmentViewModel: AppointmentViewModel
 }
 ```
 
-To continue with this example: you are next working on a user control to present the appointment card. Firstly, you must first add a reference to the view models. Next check that you have set a suitable design width and/or height. Then you can add some XAML for the design data context, as shown:
+To continue with this example: you are next working on a user control to present the appointment card. Firstly, you must first add a reference to the view models. Next check that you have set a suitable design width and/or height. Then you can add some XAML for the design data context. 
 
+As of Avalonia version 11, the default sample app is set to use Compiled bindings. This requires two declarations:
```
stevemonaco, line 66: For future reference, I'm not sure we need to point this out everywhere. At some point (very soon), compiled bindings will be ubiquitous and these asides become noise.

stevemonaco (approved):

PR #444, opened 2024-04-05
docs (FilePickerOptions) update with webp details

Nicholas-Westby:
Adds details about the recently added built-in support for WebP images.

At least, so long as this other PR I created is accepted/merged: https://github.com/AvaloniaUI/Avalonia/pull/15242

I also took the opportunity to note that the `mdls` command can be used to find new uniform type identifiers.


stevemonaco (commented):
Can merge if upstream PR is accepted.

We might need a better way to specify versions of small changes like these in docs. `<MinVersion ... />` doesn't quite fit with code blocks (and needs some adjustments of its own). This feature would make it to 11.2 if accepted and possibly 11.1.x if backported too.

maxkatz6:
I can take it into 11.1 no problem, just need to add MinVersion here.

docs/concepts/services/storage-provider/file-picker-options.md
```diff
@@ -91,6 +91,7 @@ Avalonia has set of built-in file types:
 - FilePickerFileTypes.ImageAll - all images
 - FilePickerFileTypes.ImageJpg - jpg images
 - FilePickerFileTypes.ImagePng - png images
+- FilePickerFileTypes.ImageWebP - webp images
```
stevemonaco, line 94: To clarify, we have a MinVersion tag to indicate certain features are new. In particular, `<MinVersion version="11.1" />` for your case can be used to indicate that WebP has been newly added.

You could opt to make a separate, short section if you feel that's better. ie. `## WebP Options <MinVersion version="11.1" />` to clarify both `FilePickerTypes.ImageWebP` and the `Patterns` part are new. Ensure that users won't be mistaken that WebP loading/rendering support is new, only the picker type changes.

If the styling seems janky on the tag where you put it, let me know and I will fix the React component after merging.

Nicholas-Westby: @stevemonaco I'm not sure if you're talking to me or @maxkatz6, but I'm happy to revise the PR according to the recommendations you outlined if you're talking to me.

stevemonaco: @Nicholas-Westby Yes, if you could revise that would be great.

Nicholas-Westby: @stevemonaco OK, I have added a section explaining that the WebP changes apply to 11.1+, as well as some hints on how to accomplish similar on older versions using more custom code.

If you'd like more changes, feel free to edit for expedience, or to make further recommendations to me.

stevemonaco (approved):

docs/concepts/services/storage-provider/file-picker-options.md
stevemonaco: LGTM. Thanks for the contribution.

Issue #445, opened 2024-04-05
TabControl.ContentTemplate uses wrong DataContext #2


SimonORorke:
This is a follow on from [closed Issue 13909 TabControl.ContentTemplate uses wrong DataContext](https://github.com/AvaloniaUI/Avalonia/discussions/13909), as I suggest it warrants further discussion. Let me reformulate the problem in a more long-winded way.

Following the instructions in the [How To Bind Tabs](https://docs.avaloniaui.net/docs/guides/data-binding/how-to-bind-tabs) page of Avalonia's documentation, in `MainWindow.axaml` I bound a `TabControl` to a collection property of `MainWindowViewModel`, like this:
```
<TabControl x:Name="TabControl"
			ItemsSource="{Binding Tabs}"
			SelectedItem="{Binding SelectedTab}">
	<TabControl.ItemTemplate>
		<DataTemplate>
			<TextBlock Text="{Binding Header}" />
		</DataTemplate>
	</TabControl.ItemTemplate>
	<TabControl.ContentTemplate>
		<DataTemplate>
			<ContentControl Content="{Binding ViewModel}" />
		</DataTemplate>
	</TabControl.ContentTemplate>
</TabControl>
```
The problem is apparent when looking at the same XAML in Rider:
![TabControl Binding Error](https://github.com/AvaloniaUI/Avalonia/assets/79874679/a001543d-92ef-4590-ac2c-74150c6b52c1)
`TabControl.ItemTemplate`, which defines the tab header, correctly binds to a property in the repeating item of the collection to which `ItemsSource` is bound. However, `TabControl.ContentTemplate`, which defines the content shown below the tab, incorrectly attempts to bind to a a non-existent property of the `Window`'s view model instead of the required property of the repeating item.

The solution, as I found in [Issue 13909](https://github.com/AvaloniaUI/Avalonia/discussions/13909), is to add a `DataType` to `TabControl.ContentTemplate`'s `DataTemplate`, like this:
```
<TabControl.ContentTemplate>
	<DataTemplate DataType="viewModel:TabItemViewModel">
		<ContentControl Content="{Binding ViewModel}" />
	</DataTemplate>
</TabControl.ContentTemplate>
```

Is the difference between the binding behaviours of `TabControl.ItemTemplate` and `TabControl.ContentTemplate` a bug?  Or is it as expected?

[...]

PR #446, opened 2024-04-05
Fixed TabControl.ContentTemplate's binding

SimonORorke:
As discussed in https://github.com/AvaloniaUI/avalonia-docs/issues/445.


docs/guides/data-binding/how-to-bind-tabs.md
```diff
@@ -49,7 +49,7 @@ Finally create a `TabControl` and bind its `ItemsSource` property to the DataCon
       <DataTemplate>
         <DockPanel LastChildFill="True">
           <TextBlock Text="This is content of selected tab" DockPanel.Dock="Top" FontWeight="Bold" />
-          <TextBlock Text="{Binding Content}" />
+          <TextBlock Text="{Binding Content DataType="TabItemModel"}" />
```
timunie, line 52: DataType should go into the DataTemplate-node above imo

docs/guides/data-binding/how-to-bind-tabs.md
```diff
@@ -49,7 +49,7 @@ Finally create a `TabControl` and bind its `ItemsSource` property to the DataCon
       <DataTemplate>
```
timunie, line 49: here ^

timunie (changes requested):

docs/guides/data-binding/how-to-bind-tabs.md
SimonORorke: Ah yes, that's where it goes, not enough caffeine first thing in the morning here. I'll fix it.

SimonORorke: Sorry, still wrong

docs/guides/data-binding/how-to-bind-tabs.md
SimonORorke: Sorry for beginner's confusion with pull requests. I made three pull requests for three branches on my local repository. The third one  should be right.

I got confused because, when I tried to make a correction and pull request, GitHub asked me whether I wanted to make a new branch and pull request.  I wondered whether, if I did not make a new branch, my change would not be in a pull request, so I did it to be safe.  I'm now fairly sure that was unnecessary, as GitHub had previously told me something like I can just update my original fork branch to make a pull request? (Or add to an existing pull request?)

stevemonaco: No need for new branches besides the original PR branch. Work in the same branch and make changes. What I do is commit in IDE without pushing, hop into CLI, and `git push -f origin HEAD`. The new commit will automatically propagate into your PR and be shown here.

SimonORorke: @stevemonaco, thanks for the tip. My workflow is a little different. I never commit in the IDE.  I use GitKraken for all things Git.  I wonder what would happen if I were to pushed from GitKraken rather than the Git CLI.

stevemonaco: I haven't used GitKraken, so I don't know.

SimonORorke: OK, I've successfully updated this pull request with the change @timunie suggested. (@stevemonaco, pushing from GitKraken works fine.)

I've deleted the other two branches I unnecessarily created.  For unknown reason, before I did that, I could not see pull requests for those branches.  As I will never do that again, I can put it down to one of those unexplained mysteries.  If those pull requests never existed, that would explain why they elicited no response.

timunie (approved):
NIT: one could also specify the namespace as well as call it TabItemViewModel but I think the current proposal is good enough for now.

timunie:
@SimonORorke Imo that change would be less confusing here. If you want to make that more clear you can also add a comment to the xml.

```xml
<!-- This is a comment -->
<DataTemplate DataType="vm:..." >
....
```

timunie:
LGTM! Thank you a bunch for your efforts. 🙏

PR #456, opened 2024-04-14
Update inspect-the-xaml.md

SdAm1n:
<UserControl ...
   x:Class="ToDoList.ToDoListView">

this is created instead of

<UserControl ...
   x:Class="ToDoList.Views.ToDoListView">


stevemonaco:
This is caused by a difference in whether you create the `UserControl` using the VS extension templates or the templates from https://github.com/AvaloniaUI/avalonia-dotnet-templates. The VS extension creates types that follows folder convention while avalonia-dotnet-templates seems not to. You might want to create an issue there about this and let the team figure out which approach is "right".

For now, you should rename your View (both the code-behind class and the `x:Class` in XAML) to match the tutorial.

PR #462, opened 2024-04-16
Change dark theme for code blocks to vsDark

### LINKED ISSUE #461: Dark mode syntax highlighting is mediocre
The current dark mode syntax highlighting theme uses Dracula which looks like this for the two most relevant languages (C# and XML):

image
image

Notice how there are only three colors (including white) for XML and it oddly chooses to italicize keywords in C#. The current light mode theme (GitHub) does not suffer from this issue.

I recommend that docs move to a theme that is better suited. Here is what `require('prism-react-renderer').themes.vsDark;` looks like when applied to the same page:

image

image

I can PR this if this theme change is acceptable. This only affects syntax highlighting blocks, AFAIK, and the rest of the site remains the same.

stevemonaco:
Closes #461

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -2,7 +2,7 @@
 // Note: type annotations allow type checking and IDEs autocompletion
 
 const lightCodeTheme = require('prism-react-renderer').themes.github;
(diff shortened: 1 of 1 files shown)


maxkatz6 (approved):
Looks good

Issue #460, opened 2024-04-16
XAML Using Prefix can reference a namespace in another assembly


SimonORorke:
According to the **XML Namespaces** section of the [Avalonia XAML](https://docs.avaloniaui.net/docs/basics/user-interface/introduction-to-xaml) page, 

> There is an alternative prefix using: that you can use to reference code in the current assembly.

When researching the pros and cons of the `using:` vs `clr-namespace:` prefixes for providing aliases for namespaces, I found the 2020 feature request for the `using:` syntax.  In his proposal, @maxkatz6 wrote:

> Currently Avalonia supports WPF'style namespaces in XAML files, using clr-namespace with assembly.
> With that we need to provide long path to assemblies, if namespace isn't contained in current one.
> UWP and XF supports another syntax, using "using:" keyword (for UWP it is only one available syntax).
> _With this we don't need to provide assembly, instead it will internally search for namespace in referenced projects and packages._

(My emphasis.) I happen to be working on an application where the view model is in a separate assembly from the UI, so I have checked to see whether, unlike  `clr-namespace:`, `using:` can actually provide an alias for a namespace in another assembly without specifying the assembly path.  It works.

So the documentation does not explain the full power of the `using:` prefix.  Something like this would be better:

> There is an alternative prefix `using:` that you can use to reference a namespace. When the namespace is not in the current assembly, the path of the assembly is not specified. Instead, the namespace is searched for in referenced projects and packages.

maxkatz6:
You are right, documentation isn't correct here. Also, "using:" probably should be moved up and described before "clr-namespace", as we want to keep it as a preferred approach.

SimonORorke:
I agree.  I shall have a go at fixing it up and submit a pull request.

PR #463, opened 2024-04-17
XAML Using Prefix can reference a namespace in another assembly.

SimonORorke:
As discussed in https://github.com/AvaloniaUI/avalonia-docs/issues/460.


SimonORorke:
I also changed the namespace paths in the examples. They looked to me in the original to be back to front, lower to higher in the hierarchy.  For example, I replaced `MyNamespace.AppNamespace.UI` with `AppNamespace.MyNamespace`.

stevemonaco (approved):

PR #467, opened 2024-04-24
Setting the Design-Time DataContext in Code

SimonORorke:
https://github.com/AvaloniaUI/avalonia-docs/issues/466


docs/guides/implementation-guides/ide-support.md
```diff
@@ -95,6 +95,19 @@ Alternatively you can use `Design.DataContext` attached property. As well as `De
...
 
+### Setting the Design-Time DataContext in Code
+
+Declaring the design-time `DataContext` in XAML will not work if the view model's only constructor has parameters. This happens with typical [dependency injection patterns](https://docs.avaloniaui.net/docs/guides/implementation-guides/how-to-implement-dependency-injection). One option is to use a special [design-time data model](https://docs.avaloniaui.net/docs/guides/implementation-guides/how-to-use-design-time-data) with a parameterless constructor.  However, a design-time `DataContext` that references a view model constructor with parameters may be set in code with the `Design.SetDataContext` method.
+
+```csharp
+public MainWindow()
+{
+    // This can be before or after InitializeComponent.
+    Design.SetDataContext(this, new MyViewModel(new DialogService()));
```
stevemonaco, line 106: Needs to be surrounded by an if statement - `if (Design.IsDesignMode)`. Otherwise, this is created while debugging the app and even in release mode. At least in my testing. It doesn't seem to happen if I set via XAML and I have no idea why.

docs/guides/implementation-guides/ide-support.md
```diff
@@ -95,6 +95,19 @@ Alternatively you can use `Design.DataContext` attached property. As well as `De
 </Window>
 ```
 
+### Setting the Design-Time DataContext in Code
+
+Declaring the design-time `DataContext` in XAML will not work if the view model's only constructor has parameters. This happens with typical [dependency injection patterns](https://docs.avaloniaui.net/docs/guides/implementation-guides/how-to-implement-dependency-injection). One option is to use a special [design-time data model](https://docs.avaloniaui.net/docs/guides/implementation-guides/how-to-use-design-time-data) with a parameterless constructor.  However, a design-time `DataContext` that references a view model constructor with parameters may be set in code with the `Design.SetDataContext` method.
```
stevemonaco, line 100: The first line is a bit awkward. Consider: "Setting the design-time `DataContext` in XAML directly is only possible if the ViewModel has a parameterless constructor." It's not strictly correct, but creating a custom `MarkupExtension` to use a service locator and construct the correct VM is complicated for docs.

Should change the links to be rooted or relative. eg. `[text...](/docs/guides/implementation-guides/how-to-use-design-time-data)` or '[text...](how-to-use-design-time-data)`

This perhaps should be in how-to-use-design-time-data instead, but up to you / other reviewers.

stevemonaco (changes requested):

SimonORorke:
> Needs to be surrounded by an if statement - if (Design.IsDesignMode).

Done.

>  Consider: "Setting the design-time `DataContext` in XAML directly is only possible if the ViewModel has a parameterless constructor." It's not strictly correct, but creating a custom `MarkupExtension` to use a service locator and construct the correct VM is complicated for docs.

Done. To accommodate the point that it's not strictly correct, I replaced the suggested 'directly' with 'by one of the methods shown above'.

> Should change the links to be rooted or relative

Done. I went for relative, as I can test that on my computer.

> This perhaps should be in how-to-use-design-time-data instead, but up to you / other reviewers.

My rationale for leaving the new documentation on the 'How to Use Live Preview' page: 

- Special design-time view models, the subject of how-to-use-design-time-data, are unlikely to have constructors with parameters. Rather, that approach is an _alternative_ to the technique I've documented, as the new documentation explains.
- My own experience trying to find this information. I had not considered design-time view models. So I looked on the 'How to Use Live Preview' page, as that's what I was trying to find out.

In relation to this point, I have changed the name of the view model in the example from `MyViewModel` (following other examples on the page), to `MainWindowViewModel`. This seems apt, as the technique is most likely to be used with the run-time view model rather than a special design-time view model.

stevemonaco (approved):

stevemonaco:
Changes look good. Thanks for the contribution.

PR #470, opened 2024-04-27
Fix misleading description for pseudoclass selectors

stevemonaco:
https://docs.avaloniaui.net/docs/reference/styles/style-selector-syntax#by-pseudo-class makes two incorrect statements:

* There can only be one pseudo class in the selector.
* When used in combination with other classes, the pseudo class must be the last in the list.

Changed the first and added XAML+C# usage syntax examples. Removed the second after testing both XAML+C# and finding out it wasn't true.

### DIFF
--- a/docs/reference/styles/style-selector-syntax.md
+++ b/docs/reference/styles/style-selector-syntax.md
@@ -81,18 +81,20 @@ Selects a control with the specified style class or classes. Multiple classes ar
 
 ```xml
 <Style Selector="Button:focus">
+<Style Selector="Button:focus:pointerover">
 <Style Selector="Button.large:focus">
 ```
 
 
 ```csharp title='C#'
 new Style(x => x.OfType<Button>().Class(":focus"));
+new Style(x => x.OfType<Button>().Class(":focus").Class(":pointerover"));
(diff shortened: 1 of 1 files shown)


docs/reference/styles/style-selector-syntax.md
```diff
@@ -81,18 +81,20 @@ Selects a control with the specified style class or classes. Multiple classes ar
...
 <Style Selector="Button:focus">
+<Style Selector="Button:focus:pointerover">
 <Style Selector="Button.large:focus">
 ```
 
 
 ```csharp title='C#'
 new Style(x => x.OfType<Button>().Class(":focus"));
+new Style(x => x.OfType<Button>().Class(":focus").Class(":pointerover"));
 new Style(x => x.OfType<Button>().Class("large").Class(":focus"));
```
robloo, line 92: There appears to be a missing colon `:` with `large`. Might want to fix that here too.

robloo: Nevermind... I actually read the docs. Will delete this.

maxkatz6 (approved):

PR #472, opened 2024-04-27
Add basic details on how to configure a custom log sink

stevemonaco:
Custom log sinks were a requested topic in the community.

I did not add an actual implementation as basing off of `TraceLogSink` is ~150 LoC, plus adding `StringBuilderCache`. So instead of making a trimmed down version with less efficient string building, I linked to the `TraceLogSink` source.


maxkatz6 (approved):
Yeah, it should be enough.

PR #473, opened 2024-04-28
Update reactive-view-model.md

dbchandra23:
I had great difficulty and confusion to properly use ReactiveUI with Avalonia. Due to the felxible nature of Avalonia, there was many different ways I found in the forums and even the current docs that is not really great for ReactiveUI. I have provided a simple example that works elegantly with the 11.x Avalonia and ReactiveUI.

### DIFF
--- a/docs/concepts/reactiveui/reactive-view-model.md
+++ b/docs/concepts/reactiveui/reactive-view-model.md
@@ -59,3 +59,65 @@ Any change to the view model description property is achieved using the `set` ac
 When _Avalonia UI_ uses the binding to **Update** the view model, the `set` accessor ensures that any parts of the view model that depend on the description property can also react to the change if necessary.
 
 On the next page, you will learn how a reactive command acts as a special case of the view model update.
+
+## Performing Data Validation
+ReactiveUI provides the ability to perform data validation from the ViewModel that integrates well with the Avalonia fluent controls. The example shows how this can be done. The example is for a ViewModel that exposes a directory path string that is bounded to a TextBox control on the View.
+
+Steps to enable data validation on the ViewModel:
+1. Download the ReactiveUI.Validation nuget package (for desktop applications)
+2. Change the inheritance of the ViewModel to ReactiveValidationObject
+3. Include a this.ValidationRule in the ViewModel constructor, which provides the validation check for the ViewModel property and corresponding validation error message.
+
+The code below shows the MainWindowViewModel.cs and MainWindow.axaml 
+```C#
+using System.IO;
+using ReactiveUI;
+using ReactiveUI.Validation.Extensions;
+using ReactiveUI.Validation.Helpers;
+
+namespace AvaloniaApplication3.ViewModels;
+
+public class MainWindowViewModel : ReactiveValidationObject
+{
+    private string _directoryPath = string.Empty;
+    
+    public string DirectoryPath
+    {
+        get => _directoryPath;
+        set => this.RaiseAndSetIfChanged(ref _directoryPath, value);
+    }
+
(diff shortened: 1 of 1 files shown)


docs/concepts/reactiveui/reactive-view-model.md
```diff
@@ -59,3 +59,65 @@ Any change to the view model description property is achieved using the `set` ac
 When _Avalonia UI_ uses the binding to **Update** the view model, the `set` accessor ensures that any parts of the view model that depend on the description property can also react to the change if necessary.
 
 On the next page, you will learn how a reactive command acts as a special case of the view model update.
+
+## Performing Data Validation
```
stevemonaco, line 63: I think that this should have its own page under the ReactiveUI section, especially if this is a standalone example. Can be done in a future PR though.

docs/concepts/reactiveui/reactive-view-model.md
```diff
@@ -59,3 +59,65 @@ Any change to the view model description property is achieved using the `set` ac
...
+    <TextBox
+        Name="DirectoryPathTextBox"
+        VerticalAlignment="Top"
+        Text="{Binding DirectoryPath}"/>
+
+</Window>
+```
+Below is the output:
+
+![Avalonia-ReactiveUI-DataValidation](https://github.com/AvaloniaUI/avalonia-docs/assets/130417839/75aede6b-f8cc-4bb8-8520-f9a8f24aa779)
```
stevemonaco, line 123: The image should also be committed to the repository. ie. in `/img/concepts/reactiveui`.

docs/concepts/reactiveui/reactive-view-model.md
```diff
@@ -59,3 +59,65 @@ Any change to the view model description property is achieved using the `set` ac
...
+    public string DirectoryPath
+    {
+        get => _directoryPath;
+        set => this.RaiseAndSetIfChanged(ref _directoryPath, value);
+    }
+
+    public MainWindowViewModel()
+    {
+        DirectoryPath = "C:\\BogusPath\\";
+        this.ValidationRule(
```
stevemonaco, line 93: I do wish there was more than one validation. Ideally one that requires multiple properties, but examples are hard to write.

docs/concepts/reactiveui/reactive-view-model.md
```diff
@@ -59,3 +59,65 @@ Any change to the view model description property is achieved using the `set` ac
 When _Avalonia UI_ uses the binding to **Update** the view model, the `set` accessor ensures that any parts of the view model that depend on the description property can also react to the change if necessary.
 
 On the next page, you will learn how a reactive command acts as a special case of the view model update.
+
+## Performing Data Validation
+ReactiveUI provides the ability to perform data validation from the ViewModel that integrates well with the Avalonia fluent controls. The example shows how this can be done. The example is for a ViewModel that exposes a directory path string that is bounded to a TextBox control on the View.
```
stevemonaco, line 64: A bit imprecise to say it integrates with the Avalonia Fluent controls as it will work for any theme. It would integrate with one of Avalonia's data validation plugins: either Data Annotations, `INotifyDataErrorInfo`, or via exceptions. I'm not sure which one ReactiveUI is using here.

stevemonaco (changes requested):
* I think that this should have its own page under the ReactiveUI section, especially if this is a standalone example.
* The image should also be committed to the repository. ie. in `/img/concepts/reactiveui`.
* I do wish there was more than one validation.
* A bit imprecise to say it integrates with the Avalonia Fluent controls. It would integrate with one of Avalonia's data validation plugins.

PR #477, opened 2024-05-03
Add new services documentation

maxkatz6:


docs/concepts/services/insets-manager.md
```diff
@@ -14,6 +14,10 @@ var insetsManager = TopLevel.GetTopLevel(control).InsetsManager;
 As for now, this service has implementation only on mobile and browser backends. If you need to adjust desktop window decorations, please use `Window.ExtendClientAreaToDecorationsHint`, `Window.ExtendClientAreaChromeHints`, `Window.ExtendClientAreaTitleBarHeightHint` properties.
 :::
 
+:::note
+Starting with Avalonia 11.1, any Avalonia app will automatically adjsut its root view accordingly to the inset values. This behavior can be disabled by setting `TopLevel.AutoSafeAreaPadding="False"` attached property value on the root view.
```
stevemonaco: adjsut -> adjust

docs/concepts/services/input-pane.md
```diff
@@ -0,0 +1,74 @@
...
+
+The `InputPane` allows developers to listen for the platform's input pane (e.g., software keyboard or on-screen keyboard) current state and boundaries.
+
+The `InputPane` can be access through an instance of `TopLevel` or `Window`, for more details on accessing `TopLevel` please visit [TopLevel](../toplevel) page:
+```cs
+var inputPane = TopLevel.GetTopLevel(control).InputPane;
+```
+
+:::note
+Currently, Avalonia does not* automatically adjust root view and scrolling position depending on state of the input pane. Instead, it is recommended for developers to use IInputPane API and adjust their apps accordingly.
```
stevemonaco, line 16: Was "not*" intentional or a typo / missing markup italic?
input pane -> `InputPane`, if we're using proper nouns. A few other occurences in the doc.
Need backticks around IInputPane, too.

maxkatz6: It was, since I added details below:
```markdown
\* Automatic adjustment is planned for future 11.* releases.
```
Will remove that anyway, will keep as two paragraphs instead.

maxkatz6: > input pane -> InputPane

Almost everywhere "input pane" is used, I copied from the Avalonia xml docs. I would keep it consistent in case if we auto-generate these one day.

stevemonaco (approved):

Arhell (approved):

PR #499, opened 2024-06-02
Add helpful context to example in tray-icon.md

logiclrd:
This PR updates the example in tray-icon.md in two ways:

1. The `<TrayIcon.Icons>` snippet is placed in context. When I tried to use this yesterday, I saw the snippet and thought, "Well, that's great but where do I _put_ that?" The page does say `App.xaml`, and yes, you do just drop it in there, but there was uncertainty there that a `<TrayIcon.Icons>` element could simply go directly in as a child of the root `<Application>`.

2. The sample references an icon as "/Assets/avalonia-logo.ico", but it doesn't say how you get that icon there. After some fiddling and searching and experimentation, I finally discovered that it works if I put an `<AvaloniaResource>` item into my `.csproj`. So, this PR adds a mention of this fact for future readers.

### DIFF
--- a/docs/reference/controls/detailed-reference/tray-icon.md
+++ b/docs/reference/controls/detailed-reference/tray-icon.md
@@ -28,25 +28,39 @@ You must use a **native menu** with the tray icon, and not the _Avalonia UI_ men
 This example defines a simple tray icon menu in the `App.xaml` file :
 
 ```xml
-<TrayIcon.Icons>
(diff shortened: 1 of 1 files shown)


stevemonaco (approved):
LGTM. Project templates historically included the Avalonia icon by default. I think some templates were introduced over time that don't though. Most people will use their IDE for the `AvaloniaResource` step.

PR #511, opened 2024-06-18
Update defining-properties.md

sedwick2048:
Switched out "MainWindow.axaml.cs" for "MyCustomControl.axaml.cs" to accurately reflect which code file is being displayed.


stevemonaco (changes requested):
This should be `MyCustomControl.cs` instead. There's no XAML involved for a self-rendering control like this one.

stevemonaco (approved):

stevemonaco:
Thanks for PR!

PR #519, opened 2024-07-17
Update styles-and-resources/resources.md

asmirnov82:
Add information about registering newly created resource dictionary into App.axml file


docs/guides/styles-and-resources/resources.md
```diff
@@ -108,7 +108,21 @@ To add a resource dictionary file, follow this procedure:
...
+    <ResourceDictionary.MergedDictionaries>
+      <MergeResourceInclude Source="/Assets/AppResources.axaml" />
+    </ResourceDictionary.MergedDictionaries>
+  </ResourceDictionary>
+</Application.Resources>
+```
+
+In the above example, the resources file AppResources.axaml is located in the /Assets project folder.
+
+You can now add the resources you want to define into AppResources.axaml in the position indicated. It looks like this:
```
maxkatz6, line 125: This text duplicates "Include and Merge Resources" from below.

maxkatz6: Maybe this section should be moved instead. Or linked.

asmirnov82: It's not a full duplicate as "Include and Merge Resources" section says about styles, that can be implemented in separate file from the resources and uses `Styles.Resources` xaml section (not an `Application.Resources`). As I am not an expert in Avalonia and XAML, I missed this info (decided that it's irrelevant to my case) and spent quite a lot of time trying to figure out why following this guide doesn't work for me. I believe, that I am not the only one facing with this issue.

So, what would you suggest me to do (reference or link the section or moved the section out instead)? 

Even though the presence of styles and merged dictionaries defined at the various levels of an application are described in `Merged Resources Priority` section, for me it's not clear what is the correct strategy of choosing the level where to merge the required resource dictionary file. So I would prefer this to be clarified in the guide (so just moving this section out doesn't look as a good choice from my perspective). Could you please help in achieving this?

asmirnov82: @maxkatz6 this is fixed. Please review

maxkatz6 (approved):
I just noticed it adds `MergeResourceInclude` as a documented feature now. Which isn't what I would prefer it to be, but should be fine.

Everything else is good.

PR #538, opened 2024-09-02
Update Test Drive section for consistency between VS, CLI, and Rider

mini-batch:
The Visual Studio sections of the Test Drive tutorial previously used a template which resulted in inconsistencies between the docs and a VS user's project - leading to confusion as seen in the comments regarding commit adfbe40.

These changes update the VS sections to use the _Avalonia .NET MVVM_ template, which results in the same project files as CLI and Rider, and removes the use of the Template Wizard -> increasing uniformity between the different options.


jindongjie (approved):

luke-whos-here:
Similar changes were made in the newest version of the Getting Started guide.

PR #542, opened 2024-09-18
[ios docs] Updated ios docs to mention mention additional relevant info

llfab:
Mentioned additional learnings from starting a new iOS project on a most up to date Apple Silicon Mac. 

I did not change the line where you mention to use the same bundle id but just explained in the bottom info box that the bundle Id may be changed in the actual Avalonia project


llfab:
Reverted the wrong bundle id comment which just leaves one file changed

maxkatz6 (approved):

PR #547, opened 2024-09-22
Update opening-a-dialog.md

nojuslau:
System.NullReferenceException: Object reference not set to an instance of an object."
10:09:51.696 [Error] 18512 <= "   at Avalonia.MusicStore.Views.MainWindow.<.ctor>b__0_0(Action1 action) in Avalonia.MusicStore\Views\MainWindow.axaml.cs:line 14"

Either ensure that ViewModel is not nullable or add data context in constructor: DataContext = new MainWindowViewModel();

### DIFF
--- a/docs/tutorials/music-store-app/opening-a-dialog.md
+++ b/docs/tutorials/music-store-app/opening-a-dialog.md
@@ -145,8 +145,13 @@ private async Task DoShowDialogAsync(InteractionContext<MusicStoreViewModel,
 - Add the following code to the end of the constructor:
 
 ```csharp
-this.WhenActivated(action => 
-         action(ViewModel!.ShowDialog.RegisterHandler(DoShowDialogAsync)));
+this.WhenActivated(action =>
+{
+     if (viewModel != null)
+     {
+        action(ViewModel!.ShowDialog.RegisterHandler(DoShowDialogAsync)));
+     }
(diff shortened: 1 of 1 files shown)


thevortexcloud:
The view model is nullable. The code that is there just assumes it never will be, which it wont 99% of the time. The only time it should be null is if you for some reason create the dialogue without giving it a view model.

In fact: adding this check may actually create additional bugs since it will silently fail to register interaction handlers when the view is activated.

ruxo:
> @Simusharp Are you sure about this change?
> 
>     1. This whole section is designed to teach IsNull as an intro to converters. This is not a typo-level mistake. It'd be surprising if it was _this_ wrong for _this_ long.
> 
>     2. If the IsNull in the code block is wrong, then it's probably also wrong in the paragraph that references the code block. You changed the code block but didn't address any of the corresponding tutorial text.
> 
>     3. Did you test this code change? I don't think there's any automation for testing tutorial code blocks, but it'd be good to know if you've validated the PR by running the code

I just tried the example today. Yes, it's wrong, it has to be `IsNotNull` as the OP suggested.

I've add another PR https://github.com/AvaloniaUI/avalonia-docs/commit/4828b55990de87c21ca315647dc896ac707f75c1 ,fixed as suggested.

PR #552, opened 2024-09-29
Update displaying-images.md

Simusharp:
There is a typo in the logic; it should IsVisible when IsNotNull

### DIFF
--- a/docs/tutorials/music-store-app/displaying-images.md
+++ b/docs/tutorials/music-store-app/displaying-images.md
@@ -201,7 +201,7 @@ Follow this procedure:
 - Add this data binding and converter to the panel element below:
 
 ```
-IsVisible="{Binding Cover, Converter={x:Static ObjectConverters.IsNull}}"
+IsVisible="{Binding Cover, Converter={x:Static ObjectConverters.IsNotNull}}"
 ```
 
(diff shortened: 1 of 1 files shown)


djeikyb:
@Simusharp  Are you sure about this change?

1. This whole section is designed to teach IsNull as an intro to converters. This is not a typo-level mistake. It'd be surprising if it was _this_ wrong for _this_ long.
2. If the IsNull in the code block is wrong, then it's probably also wrong in the paragraph that references the code block. You changed the code block but didn't address any of the corresponding tutorial text.
3. Did you test this code change? I don't think there's any automation for testing tutorial code blocks, but it'd be good to know if you've validated the PR running the code

ruxo:
> @Simusharp Are you sure about this change?
> 
>     1. This whole section is designed to teach IsNull as an intro to converters. This is not a typo-level mistake. It'd be surprising if it was _this_ wrong for _this_ long.
> 
>     2. If the IsNull in the code block is wrong, then it's probably also wrong in the paragraph that references the code block. You changed the code block but didn't address any of the corresponding tutorial text.
> 
>     3. Did you test this code change? I don't think there's any automation for testing tutorial code blocks, but it'd be good to know if you've validated the PR by running the code

I just tried the example today. Yes, it's wrong, it has to be `IsNotNull` as the OP suggested.

I've add a new PR, which has fixed as suggested: https://github.com/AvaloniaUI/avalonia-docs/commit/4828b55990de87c21ca315647dc896ac707f75c1

PR #554, opened 2024-09-30
Update add-data-persistence.md

Simusharp:
In a load method, if the directory does exist, we should return an empty collection not creating the directory and iterate over the files.

### DIFF
--- a/docs/tutorials/music-store-app/add-data-persistence.md
+++ b/docs/tutorials/music-store-app/add-data-persistence.md
@@ -53,7 +53,7 @@ public static async Task<IEnumerable<Album>> LoadCachedAsync()
 {
     if (!Directory.Exists("./Cache"))
     {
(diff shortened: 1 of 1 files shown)


djeikyb:
The diff adds a bug. I don't know if it crashes the app, or if the breakage is limited to the load method. I haven't run the app, I just looked three lines down in the diff.

Simusharp:
> The diff adds a bug. I don't know if it crashes the app, or if the breakage is limited to the load method. I haven't run the app, I just looked three lines down in the diff.

Can you elaborate about the bug?

djeikyb:
On review, I'm wrong about this introducing a bug. I do wonder about keeping the code in these docs and [the tutorial repo](https://github.com/AvaloniaUI/MusicStoreTutorial) in sync.

PR #556, opened 2024-10-10
Add missing info about .UseReactiveUI() call inside BuildAvaloniaApp()

MrClan:
Includes the essential step of adding `.UseReactiveUI` when utilizing Reactive features, as outlined in the documentation. Omitting this call results in a runtime exception, which may cause confusion for new users.

Exception encountered: ```System.ArgumentException: Unable to detect activation/deactivation of Project1.Views.MainWindow; implementing IActivationForViewFetcher may be required."```

### DIFF
--- a/docs/tutorials/music-store-app/opening-a-dialog.md
+++ b/docs/tutorials/music-store-app/opening-a-dialog.md
@@ -126,6 +126,17 @@ At this point, the code for the interaction is still incomplete. If you attempt 
 
 Your next step is to make sure that the main window view knows how to start the interaction. This is implemented in the code-behind file for the main window view, and uses some features of the _ReactiveUI_ framework.  Follow this procedure:
 
+- Open **Program.cs** file. Add an additional call to `UseReactiveUI()` inside **BuildAvaloniaApp** method as follows:
+
+```csharp
+    public static AppBuilder BuildAvaloniaApp()
+        => AppBuilder.Configure<App>()
+            .UsePlatformDetect()
+            .WithInterFont()
+            .LogToTrace()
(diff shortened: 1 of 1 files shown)


docs/tutorials/music-store-app/opening-a-dialog.md
```diff
@@ -126,6 +126,17 @@ At this point, the code for the interaction is still incomplete. If you attempt
 
 Your next step is to make sure that the main window view knows how to start the interaction. This is implemented in the code-behind file for the main window view, and uses some features of the _ReactiveUI_ framework.  Follow this procedure:
 
+- Open **Program.cs** file. Add an additional call to `UseReactiveUI()` inside **BuildAvaloniaApp** method as follows:
```
timunie, line 129: That is there if you had chosen RxUI in the template settings, so I think it should be only a hint for people installing ReactiveUI after creating of the project 🤔.

MrClan: You're correct.

PS:
One can follow the tutorial without issues until this step without RxUI. So, in this step of the tutorial where RxUI is used for the first time (`Interaction`, `ReactiveWindow`, etc) - some hint about re-ensuring for this config(and related RxUI packages - which are not installed by default in CT mode) would remove some confusion and help unblock newcomers(who mistakenly (like me) started with Community Toolkit).

timunie:
@MrClan just to make it clear, I didn't want to reject your modification. Instead I wanted to just convert it into a `TIP` block as it may or may not be a needed step. 

https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#admonitions

PR #577, opened 2024-11-19
Update opening-a-dialog.md 

bitwool:
In the document for the music app store in version 11.0.x, the code fails to compile. 
InteractionContext should be replaced with IInteractionContext.


Alexizx00789999 (approved):

timunie:
Could you add a note (where the interface is introduced) that this code is only valid in newer ReactiveUI which is shipped since 11.2.0? 

Thank you.

bitwool:
@timunie  I have finished adding note, but  I'm not sure if it fits your expectations.

timunie (approved):

PR #586, opened 2024-12-14
method signature issue

Razor06:
The example didn't work because it required an interface, not a class


timunie:
same as #577 . Thx for your constribution anyways 🙏, appreciate it.

PR #595, opened 2025-01-05
Revert react to version 18 while keeping the latest docusaurus

nsrahmad:
In the release notes at https://docusaurus.io/blog/releases/3.7 , they mention that they are adding early support for react 19 in preparation for docusaurus 4. They will keep supporting react 18 in 3.x series. By the time it becomes required, the rest of the ecosystem will catch up to react 19.

Screenshot_20250105_211015


MikeCodesDotNET (approved):
yolo

PR #601, opened 2025-01-10
Improve TabStrip reference page

stevemonaco:
From discussion at https://github.com/AvaloniaUI/Avalonia/discussions/17939

`TabStrip` document has numerous errors with `TabItem` being used instead of `TabStripItem`. Fleshed out a few more details along the way, such as a major use case. Removed properties section as it was incorrect anyways.

Corrected the samples / type references in the zh-Hans doc. Did not add new content. ru doc does not exist.


docs/reference/controls/tabstrip.md
```diff
@@ -9,28 +9,36 @@ import TabStripNavigation from '/img/reference/controls/tabstrip/tabstrip-naviga
 
 Displays a strip of tab headers. You can use this control as a horizontal menu.
 
-The tab strip is comprised `<TabItem>` elements. These are displayed in the sequence that they appear in the XAML. 
+The `TabStrip` is comprised of `TabStripItem` children. Children are displayed in order of appearance and can be created by either a collection of `TabStripItem`s or generated by `ItemsSource`.
 
-## Useful Property
+Unlike `TabControl`, `TabStrip` is not responsible for presenting the content of the selected tab. Instead, respond to the `SelectionChanged` event or `SelectedItem` property changes and use a separate control such as `ContentControl` to present the desired content. This allows developers full control over the displayed content, such as custom view caching.
```
nsrahmad, line 14: Instead, respond to -> Instead, `TabStrip` responds to
and use a separate -> and uses a separate
This allows developers full -> This gives developers full

docs/reference/controls/tabstrip.md
```diff
@@ -9,28 +9,36 @@ import TabStripNavigation from '/img/reference/controls/tabstrip/tabstrip-naviga
 
 Displays a strip of tab headers. You can use this control as a horizontal menu.
 
-The tab strip is comprised `<TabItem>` elements. These are displayed in the sequence that they appear in the XAML. 
+The `TabStrip` is comprised of `TabStripItem` children. Children are displayed in order of appearance and can be created by either a collection of `TabStripItem`s or generated by `ItemsSource`.
```
nsrahmad, line 12: is comprised of -> either "is composed of" or "comprises". "comprised of" is incorrect.

Children are displayed in -> `TabStripItems` are displayed in

You can use a bulleted list to improve readability like:
... can be created by:
- A collection of `TabStripItem`s
- An `ItemsSource`

nsrahmad (commented):
bunch of grammar of stylistic comments. Feel free to ignore something if you don't agree with it.

docs/reference/controls/tabstrip.md
stevemonaco: 1. `TabControl` also supports these change mechanisms, so this proposal makes it less clear. I chose to revise differently.
2. I revised this because the phrasing was awkward.
3. Better word choice, thanks.

docs/reference/controls/tabstrip.md
stevemonaco: Fixed first two points. Used an inline numbered list for the last point. A bullet point list here with separate lines breaks the reading flow too much, IMO. Thanks.

maxkatz6 (approved):

PR #607, opened 2025-01-13
[RU] Translation of the "Code-behind" page

GamerVII-NET:
Translated the "Code-behind" page into Russian while preserving the original structure and formatting. Added translations for all sections including naming conventions for `.axaml.cs` files linking XAML and code-behind through the `x:Class` attribute accessing  controls from code-behind setting control properties programmatically and handling events with examples.


maxkatz6 (approved):
LGTM

PR #609, opened 2025-01-13
Improve Styled Properties doc, Add User Control Examples

leocb:
This improves the docs regarding styled properties, specifically it puts more emphasis on the fact that the Property name convention MUST be followed. This is something I personally had problems with and could find little to no info about (`Unable to find suitable setter or adder for property` errors)

This PR also adds a page on "how to create a user control", highlighting best practices for exposing custom attributes.


luke-whos-here:
Elements of this have been included with a broader update of custom control docs in https://github.com/AvaloniaUI/avalonia-docs/pull/1075

PR #619, opened 2025-02-21
Update native-aot.md

chuongmep:
Hi, today I'm deploy with Native AOT and have issue, I spend 2 days to fix it just with add two line in config 

```cs
<ItemGroup>
    <TrimmerRootAssembly Include="Avalonia.Controls" />
    <TrimmerRootAssembly Include="Avalonia.Base" />
    <TrimmerRootAssembly Include="Avalonia.Markup.Xaml" />
</ItemGroup>
```

### DIFF
--- a/docs/deployment/native-aot.md
+++ b/docs/deployment/native-aot.md
@@ -94,6 +94,8 @@ If controls are missing at runtime:
 <ItemGroup>
     <!-- Add specific Avalonia controls you're using -->
     <TrimmerRootAssembly Include="Avalonia.Controls" />
+    <TrimmerRootAssembly Include="Avalonia.Base" />
+    <TrimmerRootAssembly Include="Avalonia.Markup.Xaml" />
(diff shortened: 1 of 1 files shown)


maxkatz6:
This document is a bit outdated. You don't need neither of these `TrimmerRootAssembly`.
If you do, you have another problem in the app. Likely using reflection bindings.

rabbitism:
For a blank avalonia app, you don't need this file at all.

chuongmep:
Thank you for your help, AOT is panic for me now.

djeikyb:
@chuongmep [I use this csproj for starting new avalonia+nativeaot apps](https://github.com/djeikyb/AvaloniaTemplates/blob/trunk/Gen3/AvaloniaApplication1/AvaloniaApplication1.csproj). It's far from the only way, but it does compile and run succesfully.

PR #623, opened 2025-02-27
Updating macOS.md

terinfire:
Updating instructions with a script that should hopefully cover folks more (needs to be run on MacOS) and explain things better.


luke-whos-here:
The current version of the macOS deployment doc covers similar topics. Some key non-overlap areas have been included in https://github.com/AvaloniaUI/avalonia-docs/pull/1019.

PR #624, opened 2025-03-04
Fix Link for "The MVVM Pattern"

Teoh-KY:
"The MVVM Pattern"  link is broken.

### DIFF
--- a/docs/basics/data/data-binding/index.md
+++ b/docs/basics/data/data-binding/index.md
@@ -37,7 +37,7 @@ Bindings can be unidirectional: where changes in the properties of a bound appli
 Binding is used with the MVVM architectural pattern, and this is one of the principle ways of programming with Avalonia UI.
 
 :::info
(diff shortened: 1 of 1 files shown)


timunie:
I can't see the broken link in the published docs. Where did you find it?

Teoh-KY:
Steps to reproduce:

1. Browse to [Data Binding](https://docs.avaloniaui.net/docs/basics/data/data-binding)
2. At the botton of the page, find "Info" section that mention "how to use the MVVM Pattern with Avalonia", left click on "[here](https://docs.avaloniaui.net/concepts/the-mvvm-pattern)."
3. You will get "Page Not Found".

Additional Info:
After i tested few times, i noticed that if right click and open with new tab, the link is correct. If left click, some how the link got changed and will get "Page Not Found". Sorry, i think the issue might be something else.

timunie (approved):

timunie:
I can now reproduce. Approved, let's test it after merge

PR #626, opened 2025-03-09
Fix the road issue in add-a-control.md

Machillka:
The "MainView.xaml" should be the "MainWindow.axaml".


luke-whos-here:
Resolved in the latest version of the getting started guide.

PR #629, opened 2025-03-14
Update macos-development.md

CastelloBrancoTecnologia:
Publish an Bundle


luke-whos-here:
Publishing macOS bundles is now covered in some detail on https://docs.avaloniaui.net/docs/deployment/macos, as well as the Plus/Pro Parcel tool.

PR #633, opened 2025-04-03
Update combobox.md for IsEditable changes

almightyju:
This is for https://github.com/AvaloniaUI/Avalonia/pull/18094 so only merge if that pull request is accepted 😄  

I'm not sure of how best to record an image for the example I've added but can happily add if someone is willing to make one?

### DIFF
--- a/docs/reference/controls/combobox.md
+++ b/docs/reference/controls/combobox.md
@@ -31,8 +31,8 @@ You will probably use these properties most often:
 | `MaxDropDownHeight`        | The maximum height for the dropdown list. This is the actual height of the list part, not the number of items that show. |
 | `ItemPanel`                | The container panel to place items in. By default, this is a StackPanel. See [this page](../../concepts/custom-itemspanel) to customise the ItemsPanel.|
 | `Styles`                   | The style that is applied to any child element of the ItemControl.                                                       |
-| `IsEditable`               | Allows the user to type any value into the `ComboBox`. Make sure to use `DisplayMemberBinding` or `TextSearch.TextBinding` if using complex item types. |
-| `Text`                     | When `IsEditable` will be the text value of the `SelectedItem` or the text a user has entered.                           |
+| `IsEditable`               | Allows the user to type any value into the `ComboBox`. Make sure to use `DisplayMemberBinding` or `TextSearch.TextBinding` if using complex item types as otherwise it will use `ToString` on the object. |
+| `Text`                     | When `IsEditable` `Text` will be the text value of the `SelectedItem` or the text a user has entered.                           |
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


docs/reference/controls/combobox.md
```diff
@@ -125,6 +127,28 @@ namespace AvaloniaControls.Views
 
 <img src={ComboBoxDataTemplateScreenshot} alt="" />
 
+When using `IsEditable` with complex types it is important to set either `DisplayMemberBinding` or `TextSearch.TextBinding`, or both.
+A complex example of the editable text using `Id` to match items when typing in text, but the showing `DisplayValue` in the dropdown is found below:
```
thevortexcloud, line 131: I think this needs to be reworded. It took me a few tries to parse it. Perhaps:

```suggestion
The below example uses a property named `Id`, to match items when typing in text to a `ComboBox` that has `IsEditable` set to true. The display value in the `ComboBox` will be the bound `DisplayValue` value: 
```

docs/reference/controls/combobox.md
```diff
@@ -125,6 +127,28 @@ namespace AvaloniaControls.Views
 
 <img src={ComboBoxDataTemplateScreenshot} alt="" />
 
+When using `IsEditable` with complex types it is important to set either `DisplayMemberBinding` or `TextSearch.TextBinding`, or both.
```
thevortexcloud, line 130: What happens if you don't?

docs/reference/controls/combobox.md
```diff
@@ -125,6 +127,28 @@ namespace AvaloniaControls.Views
...
+<ComboBox PlaceholderText="Editable" IsEditable="true"
+          ItemsSource="{Binding MyComplexItems}"
+          Text="{Binding EditableText}"
+          DisplayMemberBinding="{Binding DisplayValue}"
+          TextSearch.TextBinding="{Binding Id, DataType=viewModels:ComplexItem}"
+          SelectedItem="{Binding Selected}" />
+```
+
+```csharp title='C#'
+public record ComplexItem(int Id, string DisplayValue);
```
thevortexcloud, line 143: I generally suggest people avoid using newer C# features like records for docs. Avalonia technically supports all the way back to .NET 4.8, which means people may have to rewrite the code if they are trying to use it there.

docs/reference/controls/combobox.md
```diff
@@ -125,6 +127,28 @@ namespace AvaloniaControls.Views
 
 <img src={ComboBoxDataTemplateScreenshot} alt="" />
 
+When using `IsEditable` with complex types it is important to set either `DisplayMemberBinding` or `TextSearch.TextBinding`, or both.
+A complex example of the editable text using `Id` to match items when typing in text, but the showing `DisplayValue` in the dropdown is found below: 
+
+```xml
+<ComboBox PlaceholderText="Editable" IsEditable="true"
```
thevortexcloud, line 134: Put `IsEditable` on a different line like the rest of the properties are. It makes it a lot more readable. I almost missed it was there.

docs/reference/controls/combobox.md
almightyju: Yeah this one is tricky without a visual, I went thought a few iterations myself. Basically in that example:
* Items in the _open_ dropdown will display using `DisplayMemberBinding`
* The editable portion (i.e. `TextBox`) will select items that match `TextSearch.TextBinding` and also display the value using `TextSearch.TextBinding`

Does listing those bullet points make more sense how it works to you?

docs/reference/controls/combobox.md
almightyju: It'll throw an error

docs/reference/controls/combobox.md
almightyju: Good point, I was trying to make the c# part concise, I'll update to use a class instead

docs/reference/controls/combobox.md
thevortexcloud: It is certainly an improvement. If you do go with that, I would suggest a slight change:

Say "will select an item that matches" instead of "selects items", as it makes it sounds like it will match multiple values at once.

docs/reference/controls/combobox.md
thevortexcloud: I would say that should be mentioned somewhere then.

luke-whos-here:
A similar example was added in https://github.com/AvaloniaUI/avalonia-docs/commit/5306cf7698e5f244f21f6b00c5095b323e2b2ae0.

PR #635, opened 2025-04-05
Update main-window.mdx (docs/get-started/test-drive/main-window#the-mainwindow-content)

0xf0pfp183:
MainViewModel -> MainWindowViewModel


timunie:
MainViewModel and MainWindowViewModel both exits, depending on the template used to create an App. This is not perfect for beginners, but a left over for historical reasons. 

Instead of changing the name forth and back, you may want to add a comment via https://docusaurus.io/docs/markdown-features/admonitions

timunie (approved):
Thx, I think it is now much more clear. 👍

PR #642, opened 2025-04-08
Added commands for updating Accelerate Developer Tools.

grokys:
Hid them in a `<details>` as to separate them from the install commands.


maxkatz6 (approved):
App already shows these commands. But duplicating here is not a problem.

PR #652, opened 2025-04-24
Add packageSourceMapping examples to quickstart guides

khalidabuhakmeh:
Updated the media-player and webview quickstart documentation to include examples of packageSourceMapping.

Without the package source mapping, build tools can be confused about where to search for certain packages when central package management is used. Adding a package source mapping helps resolve any ambiguity and lets those tools build successfully.

I ran into this issue when using Cake to build using `dotnet cake`.

https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping#enabling-package-source-mapping

### DIFF
--- a/accelerate/components/media-player/quickstart.md
+++ b/accelerate/components/media-player/quickstart.md
@@ -34,7 +34,15 @@ Avalonia Accelerate packages are distributed through a dedicated NuGet feed that
       <add key="ClearTextPassword" value="YOUR_LICENSE_KEY" />
     </avalonia-accelerate>
   </packageSourceCredentials>
-// highlight-end
+  <packageSourceMapping>
+     <packageSource key="api.nuget.org">
+        <package pattern="*" />
+     </packageSource>
+     <packageSource key="avalonia-accelerate">
+        <package pattern="Avalonia.Controls.MediaPlayer" />
+     </packageSource>
+  </packageSourceMapping>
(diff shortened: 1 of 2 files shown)


maxkatz6:
Does it work fine, if you include Avalonia.Controls.DataGrid, for example? Which isn't part of this feed.

khalidabuhakmeh:
@maxkatz6 You're right, I didn't realize there were other packages that use the same prefix.

```xml
  <packageSourceMapping>
    <!-- key value for <packageSource> should match key values from <packageSources> element -->
    <packageSource key="avalonia-accelerate">
      <package pattern="Avalonia.Controls.MediaPlayer" />
      <package pattern="Avalonia.Controls.WebView" />
    </packageSource>
    <packageSource key="nuget.org">
      <package pattern="*" />
    </packageSource>
  </packageSourceMapping>
```

The package source pattern has to be more explicit then. Not sure if this is good for the docs as it would need to be updated as packages go into and out of the accelerate feed.

What are your thoughts?

maxkatz6:
@khalidabuhakmeh we are going to eventually move these new packages to the public nuget, since our licensing is done via msbuild task. So, this problem is going to be solved automatically.

Meanwhile, explicit pattern is probably fine as a temporary workaround.

maxkatz6 (approved):
Thanks!

PR #661, opened 2025-05-17
Update how-to-bind-tabs.md

Reidea:
Fixed missing `ItemsSource` binding. The binding should be a property that accesses the items collection. Renamed it to `Items`.
The class name `TabViewModel` is misleading. Renamed to `TabItemTemplate`.

### DIFF
--- a/docs/guides/data-binding/how-to-bind-tabs.md
+++ b/docs/guides/data-binding/how-to-bind-tabs.md
@@ -6,20 +6,20 @@ title: How To Bind Tabs
 
 ## Binding Support Example
 
-You can dynamically create tab items with **data binding**. To do this, bind the `ItemsSource` property of a tab control to an array of objects representing the tab header and content.
+You can dynamically create tab items with **data binding**. To do this, bind the `ItemsSource` property of a tab control to a collection of objects representing the tab header and content.
 
 You can then use a **data template** to display the objects.
 
-This example uses an array of objects created from this `TabItemViewModel` class:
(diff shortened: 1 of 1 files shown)


thevortexcloud:
> The class name TabViewModel is misleading. Renamed to TabItemTemplate.

How is it misleading? It _is_ a view model? It's not a template. The template is defined in the view using the `<DataTemplate>`. If anything this change makes it misleading/wrong.

Reidea:
> > The class name TabViewModel is misleading. Renamed to TabItemTemplate.
> 
> How is it misleading? It _is_ a view model? It's not a template.
Looks like you are right. In my understanding TabItemViewModel should always have TabItemView.axaml and is created only in such a bundle. This confuses me.

thevortexcloud:
> In my understanding TabItemViewModel should always have TabItemView.axaml

Generally, yes. But there is no requirement for that. The view model can be used with _any_ view that know how to handle it. There is nothing stopping you from instead hooking the tab item up to something like a `HeaderedContentControl` in one view and a tab control in another. This is one of the core principles of MVVM. Essentially the view is coupled to the view model but the view model is not coupled to the view.

timunie:
I still think there are good parts in this PR that I want to merge. What if we rename `TabItemVieModel` just to `ItemViewModel`? Other than that, I like the changes as they are closer to a real world usecase. 

What do you think, @Reidea ?

timunie (approved):
LGTM. Thank you for your contribution and patiance. Very much appreciate it.

PR #683, opened 2025-07-08
Show last updated date and author

luke-whos-here:
This change gets Docusaurus to display the last updated date and last updated author on each doc. It would make it easier for viewers to quickly tell how up-to-date the doc is, without needing to go into the repo to check the version history of the source file.


grokys (approved):
Thank you, LGTM!

PR #689, opened 2025-07-21
Changing all reference links to new api-docs side

timunie:
I hope there are not too many copy/pasta issues since this was quite a lot of links to change. 

I also updated the translations, so there should be no more link to the old reference docs

### DIFF
--- a/docs/reference/controls/itemsrepeater.md
+++ b/docs/reference/controls/itemsrepeater.md
@@ -127,9 +127,5 @@ The items display horizontally, and those too far to the right would be hidden i
 ## More Information
 
 :::info
-For the complete API documentation about this control, see [here](TODO).
-:::
-
-:::info
 View the source code on _GitHub_ [`ItemsRepeater.cs`](https://github.com/AvaloniaUI/Avalonia.Controls.ItemsRepeater/blob/main/src/Avalonia.Controls.ItemsRepeater/Controls/ItemsRepeater.cs)
 :::


docs/guides/development-guides/accessing-the-ui-thread.md
```diff
@@ -26,7 +26,7 @@ Use `InvokeAsync` when you need to wait for the result, and potentially want to
 Both of the above methods have a dispatcher priority parameter. You can use this with the `DispatcherPriority` enumeration to specify the queue priority that the given job should be given.
 
 :::info
-For the possible values of the `DispatcherPriority` enumeration, see [here](http://reference.avaloniaui.net/api/Avalonia.Threading/DispatcherPriority/).
+For the possible values of the `DispatcherPriority` enumeration, see [here](https://api-docs.avaloniaui.net/docs/T_Avalonia_Threading_DispatcherPriority#fields).
```
timunie, line 29: technically this is no longer an enum. Should we correct the text to something else?

docs/reference/controls/itemsrepeater.md
```diff
@@ -127,7 +127,7 @@ The items display horizontally, and those too far to the right would be hidden i
 ## More Information
 
 :::info
-For the complete API documentation about this control, see [here](http://reference.avaloniaui.net/api/Avalonia.Controls/ItemsRepeater/).
+For the complete API documentation about this control, see [here](TODO).
```
timunie: How should we deal with this? The ItemsRepeater is no longer maintained. I'd prefer to just delete this section and add a warning badge at the top. Source could be linked to new repo however.

timunie (commented):
Added two remarks for things we need to clarify before merging it.

docs/guides/development-guides/accessing-the-ui-thread.md
maxkatz6: Well, it's a special type that is used as an enum. I think it's okay

docs/reference/controls/itemsrepeater.md
maxkatz6: Yeah, let's just remove this paragraph

maxkatz6 (approved):
Did a quick check. Aside from the items repeater part, looks good.
There are some new APIs that should be referenced on the main docs too. But can be done at some point later

PR #715, opened 2025-08-28
Fix markdown syntax for note block in resources.md

mrlacey:
Similar to #714  - I should have finished reading the page before submitting the last PR ;)


timunie (approved):
thx 🙏

PR #719, opened 2025-09-04
Fixed outdated font binding examples.

hukela:
In the original example:
The `GetInstalledFontFamilyNames` method has been removed.
The `Items` property of `ComboBox` has been set to read-only and cannot be directly assigned.
Therefore, I have rewritten this example.


hukela:
I was wondering about this earlier, thanks for providing the reason. The Window-tag has now been deleted. I'm not sure if I should directly add a commit? Or should I create a new merge request to combine both changes in a single commit?

PR #746, opened 2025-10-27
Corrected documentation of FuncValueConverter

zacfromaustinpowder:
[This Avalonia PR](https://github.com/AvaloniaUI/Avalonia/pull/19920) must be merged before this PR is entirely accurate.


luke-whos-here:
My bad, folks. I just looked through #19920 again and propose a partial rollback to restore the description in line 149 about not needing to convert back.

PR #753, opened 2025-11-06
Update resources.md to fix bug and inconsistencies

Miniontoby:
Changes:

- Fixed a bug where it said `</Application.Resources>` as starting tag
- Fixed some inconsistent space usage (e.g.: no starting space, 4 spaces instead of 2)

### DIFF
--- a/docs/guides/styles-and-resources/resources.md
+++ b/docs/guides/styles-and-resources/resources.md
@@ -77,16 +77,16 @@ It is also possible to define resources for specific theme variants: Dark, Light
 
 ```xml
 <ResourceDictionary>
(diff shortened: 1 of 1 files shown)


docs/guides/styles-and-resources/resources.md
```diff
@@ -290,7 +290,7 @@ Avalonia provides different options to access Resources from code.
 ```cs
 // In this sample we have defined the resource in App.axaml and we want to look up the value in the MainWindow constructor.
 //
-//    </Application.Resources>
+//    <Application.Resources>
```
Miniontoby, line 293: This here was the main reason I started this edit, and then I noticed the other inconsistencies

PR #754, opened 2025-11-06
Update compiled-bindings.md to fix indentation last example

Miniontoby:
Updated compiled-bindings.md to fix indentation of the last example


luke-whos-here:
Hey @Miniontoby, thanks for your contributions to the docs! I happen to agree with you - this example does look neater with the indentations.

Issue #755, opened 2025-11-06
In page "Improving Performance", are the information still correct?


cdytoby:
This is the page:

https://docs.avaloniaui.net/docs/guides/development-guides/improving-performance

- TreeDataGrid is no longer publicly maintained, so all the mentions should be moved to "accelerate" documentation.
- If DataGrid is still bad for displaying a large amount of data, and I don't need to edit them, what should I use?
- How bad is DataGrid? is it suitable for displaying 1000 rows with 4 columns on modern PC? is it suitable for displaying 100000 rows with 4 columns on modern PC?

PR #756, opened 2025-11-10
Update TreeDataGrid documentation with notices and crosslinks to Accelerate

luke-whos-here:
Added callout boxes to pages in the Avalonia OSS docs where TreeDataGrid is mentioned. Purpose is to clarify that TreeDataGrid is still recommended despite the move to Accelerate, and to crosslink to Accelerate docs.

From issue https://github.com/AvaloniaUI/avalonia-docs/issues/755


luke-whos-here:
@maxkatz6 If these changes look good to you, I'll implement them on the translated pages too before I merge.

PR #775, opened 2025-11-20
Add some examples of using AutoCompleteBox with objects

MohamedA-Ibrahim:
Adds more examples of using AutoCompleteBox with objects and how to customize display and filter for them

### DIFF
--- a/docs/reference/controls/autocompletebox.md
+++ b/docs/reference/controls/autocompletebox.md
@@ -65,7 +65,7 @@ namespace AvaloniaControls.Views
 ### Using AutoCompleteBox with Objects
 When working with complex objects instead of simple strings, specify which property should be shown and how the control should filter the underlying data. The sections below cover display binding, custom filtering, and formatting the presented text.
 
-#### Displaying Objects Using ValueMemberBinding
+#### Filtering Objects Using ValueMemberBinding
 ValueMemberBinding tells the control which property of your object is displayed in the text box and used for built-in filtering.
 
 ```xml
@@ -212,13 +212,17 @@ public partial class MainWindow : Window
 
 ```
 
-#### Customizing Display Text Using a Value Converter
-You can format the text shown in the drop-down using a value converter. This is useful when you want to show multiple fields (for example, name and price).
+#### Customizing Text Using a Value Converter
+You can control the text displayed in the drop-down by using a value converter. This is helpful when you want to show a combination of fields—for example, a product’s name and price. Keep in mind that this also affects how filtering works: if you display both name and price, the control will allow filtering by either value.
+
+If you only want to change how the item appears visually (without affecting filtering), use the standard ItemTemplate property instead.
 
 ```xml
 <Window xmlns="https://github.com/avaloniaui"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:converters="using:YourNamespace.Converters">
+        xmlns:converters="using:YourNamespace.Converters"
+        xmlns:vm="clr-namespace:YourNamespace.ViewModels;assembly=YourNamespace"
+        x:DataType="vm:MainViewModel">
 
     <Window.Resources>
         <converters:ProductDisplayConverter x:Key="ProductDisplayConverter" />


docs/reference/controls/autocompletebox.md
```diff
@@ -62,6 +62,206 @@ namespace AvaloniaControls.Views
...
+        {
+            if (value is Product product)
+                return $"{product.Name} (${product.Price:F2})";
+
+            return value?.ToString();
+        }
+
+        public object? ConvertBack(object? value, Type targetType,
+                                   object? parameter, CultureInfo culture)
+            => throw new NotImplementedException();
```
luke-whos-here: Presumably ConvertBack is never going to be needed in this case. Should this result in a DoNothing or similar, instead of throwing an exception?

docs/reference/controls/autocompletebox.md
```diff
@@ -62,6 +62,206 @@ namespace AvaloniaControls.Views
...
+
+        // Custom filter searching both Name and Id
+        ProductAutoComplete.ItemFilter = (search, item) =>
+        {
+            if (item is Product product && !string.IsNullOrWhiteSpace(search))
+            {
+                return product.Name.Contains(search, StringComparison.OrdinalIgnoreCase)
+                    || product.Id.ToString().Contains(search);
+            }
+            return true;
```
luke-whos-here, line 208: Should this be returning `false` when the input is null?

luke-whos-here:
@MrJul Can I get you to have a look at these samples? I tested them and they seem functional. Just want to get your okay that they're consistent with our general recommendations for Avalonia.

docs/reference/controls/autocompletebox.md
timunie: imo it should throw `NotSupportedException`

docs/reference/controls/autocompletebox.md
MohamedA-Ibrahim: If input is null that means there is no filter applied so we should return all entries

docs/reference/controls/autocompletebox.md
MohamedA-Ibrahim: @luke-whos-here @timunie Modified as suggested

docs/reference/controls/autocompletebox.md
luke-whos-here: Makes sense, thanks.

docs/reference/controls/autocompletebox.md
```diff
@@ -62,6 +62,206 @@ namespace AvaloniaControls.Views
...
+    }
+}
+
+```
+
+#### Customizing Display Text Using a Value Converter
+You can format the text shown in the drop-down using a value converter. This is useful when you want to show multiple fields (for example, name and price).
+
+```xml
+<Window xmlns="https://github.com/avaloniaui"
```
MrJul, line 221: `x:DataType="vm:MainViewModel"` should be specified here, otherwise the following example won't build with compiled bindings enabled (and those are enabled by default in our templates).

docs/reference/controls/autocompletebox.md
```diff
@@ -62,6 +62,206 @@ namespace AvaloniaControls.Views
...
+            }
+            return true;
+        };
+    }
+}
+
+```
+
+#### Customizing Display Text Using a Value Converter
+You can format the text shown in the drop-down using a value converter. This is useful when you want to show multiple fields (for example, name and price).
```
MrJul: This is not entirely correct. `ValueMemberBinding` doesn't just change the display, but what value is used for searching (as correctly mentioned in the section above). For example, the price will be included in the search in the example below, which might be undesirable.

This should be made clear in this sample and its title.

If the user wants to change the display only, they should use the standard `ItemTemplate` property instead.

docs/reference/controls/autocompletebox.md
```diff
@@ -62,6 +62,206 @@ namespace AvaloniaControls.Views
 
 <img src={AutoCompleteBoxScreenshot} alt="" />
 
+### Using AutoCompleteBox with Objects
+When working with complex objects instead of simple strings, specify which property should be shown and how the control should filter the underlying data. The sections below cover display binding, custom filtering, and formatting the presented text.
+
+#### Displaying Objects Using ValueMemberBinding
```
MrJul: As you wrote just below, `ValueMemberBinding` doesn't just affect display, but filtering, which is the main purpose of this property.

```suggestion
#### Filtering Objects Using ValueMemberBinding
```

docs/reference/controls/autocompletebox.md
MohamedA-Ibrahim: Added

docs/reference/controls/autocompletebox.md
MohamedA-Ibrahim: Modified to resolve the incorrect info

MrJul:
> @MrJul Modified the instructions per your comments. can you check?

The changes look good to me, thank you!

PR #883, opened 2026-02-24
Compiled-Bindings: Replace the dead codeproject link

F0bes:
Replaces the broken codeproject link with a working one.

Alternatively, we could link to an archive: https://web.archive.org/web/20240613021450/https://www.codeproject.com/Articles/1161127/Why-is-reflection-slow

It's a shame that they broke probably all of their links.


luke-whos-here (approved):
Confirmed the updated link works. Thanks!

PR #894, opened 2026-03-03
Document latest v12 breaking changes

MrJul:
Updates the Avalonia v12 breaking changes document with the changes from the latest pull requests.


MrJul:
I've just added a missing section, about the gestures being moved.

Issue #888, opened 2026-02-26
Document FileActivatedEventArgs


MrJul:
In the [Activatable Lifetime](https://github.com/AvaloniaUI/avalonia-docs/blob/main/docs/concepts/services/activatable-lifetime.md) documentation, `ProtocolActivatedEventArgs` (corresponding to `ActivationKind.OpenUri)` is documented, but `FileActivatedEventArgs` (corresponding to `ActivationKind.File`) isn't.

Reading the documentation may leave the user thinking that `ProtocolActivatedEventArgs` is all that's needed.

PR #895, opened 2026-03-04
V12 activatable lifetime edits

luke-whos-here:
General edits to improve Activatable Lifetimes doc, and to add a new section on FileActivatedEventArgs as requested in https://github.com/AvaloniaUI/avalonia-docs/issues/888

### DIFF
--- a/reference/services/activatable-lifetime.md
+++ b/reference/services/activatable-lifetime.md
@@ -42,7 +42,7 @@ Returns **true** if it was possible on the given platform. Otherwise, returns **
 You may want an app to pause or stop some code processing when it is in the background, e.g., pausing multimedia playback, or disabling recurrent HTTP requests.
 
 ```csharp
-if (Application.Current.TryGetFeature<IActivatableLifetime>() is { } activatableLifetime)
+if (Application.Current?.TryGetFeature<IActivatableLifetime>() is { } activatableLifetime)
 {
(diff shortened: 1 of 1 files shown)


reference/services/activatable-lifetime.md
```diff
@@ -85,20 +81,54 @@ if (Application.Current.TryGetFeature<IActivatableLifetime>() is { } activatable
...
+:::
+
+### Handling file activation
+
+Your app may need to handle file activation, which occurs when the OS launches or foregrounds your app (usually because the user opens a file associated with it). Like link schemas, file type associations must be registered in the system and linked to your app. Once registered, opening an associated file also opens your app via this event.
+
+Typical use cases are opening a document, importing a file, or processing files passed from the OS shell.
+
+```csharp
+if (Application.Current.TryGetFeature<IActivatableLifetime>() is { } activatableLifetime)
```
MrJul: Note: `Application.Current` is nullable, and even if it's very likely not null at the usage point, let's have code snippets that don't produce warnings by default.

```suggestion
if (Application.Current?.TryGetFeature<IActivatableLifetime>() is { } activatableLifetime)
```

(The existing code snippet above should also be adjusted.)

MrJul:
> Please have a look and let me know what you think.

That's great, thank you!
Using a more direct style is also good.

I only left a minor comment on the code.

PR #915, opened 2026-03-17
Tweaked "bindings from code" doc.

grokys:
- Move the most important info to the top: creating a compiled binding and binding it.
- Add links to API reference
- Add note about observable binding lifetimes
- General tidying

### DIFF
--- a/docs/data-binding/binding-from-code.md
+++ b/docs/data-binding/binding-from-code.md
@@ -1,9 +1,8 @@
 ---
 id: binding-from-code
 title: How to bind from code
-description: Create and apply data bindings programmatically in C# instead of XAML.
 doc-type: how-to
-description: This guide shows you how to create and manage data bindings from C# code rather than XAML.
+description: Create and manage data bindings from C# code rather than XAML.
 ---
 
 ## Creating compiled bindings from code
--- a/docs/data-binding/binding-from-code.md
+++ b/docs/data-binding/binding-from-code.md
@@ -5,12 +5,63 @@ description: Create and apply data bindings programmatically in C# instead of XA
 doc-type: how-to
 ---
 
+## Creating compiled bindings from code
 
-Binding from code in Avalonia works somewhat differently to WPF/UWP. At the low level, Avalonia's binding system is based on Reactive Extensions' `IObservable` which is then built upon by XAML bindings (which can also be instantiated in code).
+The [`CompiledBinding.Create`](/api/avalonia/data/compiledbinding#create-method) method lets you create type-safe bindings using LINQ expressions. The expression is validated at compile time, so errors in property names produce compiler errors rather than silent runtime failures.
 
-## Subscribing to changes to a property
+The method takes two generic parameters: the type of the [`DataContext`](data-context) and the type of the property being selected. It then accepts a lambda expression to select the property to bind.
 
-You can subscribe to changes on a property by calling the `GetObservable` method. This returns an `IObservable<T>` which can be used to listen for changes to the property:
+For example, if you have a control whose [`DataContext`](data-context) is an instance of `MyViewModel`, and you want to select a `string Title { get; set; }` property from it, you would write:
+
+```csharp
+var binding = CompiledBinding.Create<MyViewModel, string>(x => x.Title);
+```
+
+Then to bind this to a control:
+
+```csharp
+textBlock.Bind(TextBlock.TextProperty, binding);
+```
+
+The [`CompiledBinding.Create`](/api/avalonia/data/compiledbinding#create-method) method takes various optional parameters to control the binding. 
+
+The following example selects a one-way binding from an explicit `viewModel` source (instead of using the data context):
+
+```csharp
+var binding = CompiledBinding.Create<MyViewModel, string>(
+    expression: vm => vm.Title,
+    source: viewModel,
+    mode: BindingMode.OneWay);
+```
+
+The expression supports nested properties, indexers, and casts:
+
+```csharp
+// Nested property
+CompiledBinding.Create<MyViewModel, string>(vm => vm.Address.City);
+
+// With a value converter
+CompiledBinding.Create<MyViewModel, bool>(
+    expression: vm => vm.IsActive,
+    converter: new BoolToOpacityConverter(),
+    mode: BindingMode.OneWay);
+```
+
+You can also use it in object initializers:
+
+```csharp
+var textBlock = new TextBlock
+{
+    [!TextBlock.TextProperty] = CompiledBinding.Create<MyViewModel, string>(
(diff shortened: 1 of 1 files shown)


docs/data-binding/binding-from-code.md
```diff
@@ -33,9 +84,21 @@ When the returned observable is subscribed, it will return the current value of
 var text = textBlock.GetObservable(TextBlock.TextProperty).Skip(1);
 ```
 
+Alternatively, you can subscribe to the [`AvaloniaObject.PropertyChanged`](/api/avalonia/avaloniaobject#propertychanged-event) event, which fires whenever _any_ property changes on the element.
+
+```csharp
+textBlock.PropertyChanged += (s, e) =>
```
grokys, line 90: We didn't mention this anywhere previously (though we mentioned much more advanced stuff).

docs/data-binding/binding-from-code.md
```diff
@@ -66,112 +133,27 @@ var textBlock = new TextBlock
 {
     Foreground = Brushes.Red,
     MaxWidth = 200,
-    [!TextBlock.TextProperty] = source.ToBinding(),
+    [!TextBlock.TextProperty] = CompiledBinding.Create<MyViewModel, string>(x => x.Title),
```
grokys, line 136: Use a compiled binding here instead of binding to an observable.

docs/data-binding/binding-from-code.md
```diff
@@ -66,112 +133,27 @@ var textBlock = new TextBlock
...
 };
 ```
 
-Using this method you can also easily bind a property on one control to a property on another:
```
grokys, line 73: Removed this - it'll cause leaks if you're not careful and is bad advice.

docs/data-binding/binding-from-code.md
```diff
@@ -66,112 +133,27 @@ var textBlock = new TextBlock
...
-Of course the indexer can be used outside object initializers too:
+The indexer can be used outside of object initializers too:
 
 ```csharp
 textBlock2[!TextBlock.TextProperty] = textBlock1[!TextBlock.TextProperty];
 ```
 
 The only downside of this syntax is that no `IDisposable` is returned. If you need to manually terminate the binding then you should use the `Bind` method.
 
-## Transforming binding values
```
grokys, line 93: This is general Rx documentation, not really relevant here - especially as we're trying to nudge people away from Rx due to lifetime issues.

docs/data-binding/binding-from-code.md
```diff
@@ -66,112 +133,27 @@ var textBlock = new TextBlock
...
-{
-    Foreground = Brushes.Red,
-    MaxWidth = 200,
-    [!TextBlock.TextProperty] = source.Select(x => "Hello " + x).ToBinding(),
-};
-```
-
 ## Using reflection bindings from code
 
-Sometimes when you want the additional features that XAML bindings provide, it's easier to use bindings from code. For example, using only observables you could bind to a property on `DataContext` like this:
```
grokys, line 109: Mots of this info was redundant. Just explain how to create a reflection binding if you really need to (you should be using compiled bindings in most cases).

docs/data-binding/binding-from-code.md
```diff
@@ -66,112 +133,27 @@ var textBlock = new TextBlock
...
-var subscription = textBlock.Bind(TextBlock.TextProperty, new ReflectionBinding("Name"));
-
-subscription.Dispose();
-```
-
-For type-safe bindings that are validated at compile time, prefer `CompiledBinding.Create` (see below).
-
-## Compiled bindings from code
-
-The `CompiledBinding.Create` factory method lets you create type-safe bindings using LINQ expressions instead of string-based property paths. The expression is validated at compile time, so typos in property names produce compiler errors rather than silent runtime failures.
```
grokys, line 141: Moved this to the top! What was the most important info doing lurking down here!?!

docs/data-binding/binding-from-code.md
```diff
@@ -197,22 +179,6 @@ private static void FooChanged(MyControl sender, AvaloniaPropertyChangedEventArg
 }
 ```
 
-## Binding to `INotifyPropertyChanged` objects
```
grokys, line 200: This is just restating what's already been said.

docs/data-binding/binding-from-code.md
```diff
@@ -56,6 +119,10 @@ subscription.Dispose();
 
 Notice that the `Bind` method returns an `IDisposable` which can be used to terminate the binding. If you never call this, then then binding will automatically terminate when the observable finishes via `OnCompleted` or `OnError`.
 
+:::note
+Unlike standard Avalonia bindings, observables to not use weak references, so you are responsible for controlling their lifetime and preventing leaks.
```
luke-whos-here:
```suggestion
Unlike standard Avalonia bindings, observables do not use weak references, so you are responsible for controlling their lifetime and preventing leaks.
```

docs/data-binding/binding-from-code.md
```diff
@@ -56,6 +119,10 @@ subscription.Dispose();
 
 Notice that the `Bind` method returns an `IDisposable` which can be used to terminate the binding. If you never call this, then then binding will automatically terminate when the observable finishes via `OnCompleted` or `OnError`.
```
luke-whos-here:
```suggestion
Notice that the `Bind` method returns an `IDisposable` which can be used to terminate the binding. If you never call this, then the binding will automatically terminate when the observable finishes via `OnCompleted` or `OnError`.
```

docs/data-binding/binding-from-code.md
```diff
@@ -5,12 +5,63 @@ description: Create and apply data bindings programmatically in C# instead of XA
 doc-type: how-to
 ---
```
luke-whos-here, line 7:
```suggestion

<p>{frontMatter.description}</p>

```

docs/data-binding/binding-from-code.md
```diff
@@ -5,12 +5,63 @@ description: Create and apply data bindings programmatically in C# instead of XA
...
```
luke-whos-here, line 6:
```suggestion
description: This guide shows you how to create and manage data bindings from C# code rather than XAML.
---
```

docs/data-binding/binding-from-code.md
luke-whos-here: Optional - could display the description as a brief intro at the top of the page, if we think it would be a helpful summary of the purpose of the page.

luke-whos-here (commented):
I'm amazed at how much better this doc is than before! I just picked up two grammar issues, and suggested adding a description which could optionally be used as an intro paragraph.

luke-whos-here (approved):

PR #948, opened 2026-04-08
947 example code on the data validation page no longer compiles with avalonia 1200

luke-whos-here:
Rewrite Data Validation page to bring content in line with v12 changes.

- Specify that DataAnnotationsValidationPlugin is disabled by default.
- Provide instructions to enable DataAnnotationsValidationPlugin.
- Remove section on INotifyDataErrorInfo validation, as the plugin is no longer available.
- Remove section on Exception validation, as the plugin is no longer available.
- Remove section on managing validation plugins, as DataAnnotationsValidationPlugin is now disabled by default anyway.
- Clarify that sample depends on ReactiveUI.
- Minor grammar and style edits.

Updating the Data Validation sample at https://github.com/AvaloniaUI/Avalonia.Samples/tree/main/src/Avalonia.Samples/MVVM/ValidationSample is out of scope of this PR and will be handled as a separate task.

### DIFF
--- a/docs/app-development/data-validation.md
+++ b/docs/app-development/data-validation.md
@@ -15,25 +15,16 @@ You can validate built-in validation attributes, [`CustomValidationAttribute`](h
 
 ### Enable `DataAnnotationsValidationPlugin`
 
-Since Avalonia v12, the data annotations validation plugin is disabled by default. To enable the plugin, perform the following steps:
+Since Avalonia v12, the data annotations validation plugin is disabled by default. To enable the plugin:
 
-1. In the `App.axaml.cs` file of your project, delete `DisableAvaloniaDataAnnotationValidation();` from the initialization completion method.
-2. In the same file, delete the entirety of the `DisableAvaloniaDataAnnotationValidation()` method, as displayed below.
+1. Go to the `Program.cs` file of your project.
+2. Within the `AppBuilder`, add this line:
 
 ```csharp
-    private void DisableAvaloniaDataAnnotationValidation()
-    {
-        // Get an array of plugins to remove
-        var dataValidationPluginsToRemove =
-            BindingPlugins.DataValidators.OfType<DataAnnotationsValidationPlugin>().ToArray();
-
-        // remove each entry found
-        foreach (var plugin in dataValidationPluginsToRemove)
-        {
-            BindingPlugins.DataValidators.Remove(plugin);
-        }
-    }
+.WithDataAnnotationsValidation()
 ```
+
+If you are migrating an earlier project that manually specified the removal of the data annotations validation plugin with `BindingPlugins.DataValidators.Remove(plugin);`, you can now delete this configuration from your initialization completion method. It is no longer needed.
 
 ### Example: The property `EMail` is required and must be a valid e-mail-address
 
--- a/docs/app-development/data-validation.md
+++ b/docs/app-development/data-validation.md
@@ -1,25 +1,27 @@
 ---
 id: data-validation
 title: Data validation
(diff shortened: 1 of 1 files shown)


docs/app-development/data-validation.md
```diff
@@ -1,81 +1,63 @@
...
 ---
 
-Avalonia offers different data validation options. This section explains how you can validate the `Properties` of your `ViewModel` and how you can style the displayed error message.
+This page explains how to use Avalonia to validate data input by users.
 
-## Validating a property
+## Data annotations validation plugin
 
-Avalonia uses [`DataValidationPlugins`](https://api-docs.avaloniaui.net/docs/T_Avalonia_Data_Core_Plugins_IDataValidationPlugin) to validate the `Properties` you bound to. Out of the box Avalonia provide these three validation plugins:
+Avalonia's `DataAnnotationsValidationPlugin` allows you to validate any [`Validation-Attributes`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute) associated with the `Properties` of your `ViewModel`.
```
grokys: Maybe just call this "Avalonia's data annotations validation support" rather than referencing the plugin by name as the name should no longer really be relevant?

docs/app-development/data-validation.md
```diff
@@ -1,81 +1,63 @@
...
+You can validate built-in validation attributes, [`CustomValidationAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.customvalidationattribute), or your own attributes derived from `ValidationAttribute`.
 
-### DataAnnotations - ValidationPlugin
+### Enable `DataAnnotationsValidationPlugin`
 
-You can decorate the `Properties` of your `ViewModel` with different [`Validation-Attributes`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute). You can use the build-in ones, use the [`CustomValidationAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.customvalidationattribute) or create your own by derive from `ValidationAttribute`.
+Since Avalonia v12, the data annotations validation plugin is disabled by default. To enable the plugin, perform the following steps:
 
-**Sample: The property EMail is required and must be a valid e-mail-address**
+1. In the `App.axaml.cs` file of your project, delete `DisableAvaloniaDataAnnotationValidation();` from the initialization completion method.
```
grokys: This is only needed if upgrading an Avalonia 11 project which had it disabled. This _doesn't_ actually enable the feature though - to enable it you also need to use `WithDataAnnotationsValidation()` on the app builder (e.g. in Program.cs):

```
        public static AppBuilder BuildAvaloniaApp()
            => AppBuilder.Configure<App>()
                .UsePlatformDetect()
                .WithInterFont()
                .LogToTrace()
                .WithDataAnnotationsValidation();
```

docs/app-development/data-validation.md
```diff
@@ -1,81 +1,63 @@
...
+2. In the same file, delete the entirety of the `DisableAvaloniaDataAnnotationValidation()` method, as displayed below.
```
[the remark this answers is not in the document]
grokys: Exactly: the `DisableAvaloniaDataAnnotationValidation` will only be present if the user added it, and added it with this name. The docs telling users how to disable it _don't_ use that name in fact: https://docs.avaloniaui.net/docs/app-development/data-validation#manage-validationplugins. I think this method naming comes from https://github.com/AvaloniaUI/avalonia-dotnet-templates/issues/262 which is just a suggestion from a user?

grokys (changes requested):

PR #950, opened 2026-04-09
Fix name of classes for Avalonia12

kant2002:


luke-whos-here (approved):
Verified that all proposed updates are correct.

PR #959, opened 2026-04-21
Rename assets file

luke-whos-here:
Rename assets.md to see if this prevents it from being excluded in website build. Update all relevant links in sidebars and other pages.


MikeCodesDotNET:
Is this a yolo test or have you confirmed it works?

luke-whos-here:
Renaming the file alone won't break the website. I've tested this on a local build and it builds and serves correctly.

I'm also fairly confident the rename will stop docs/fundamentals/assets from constantly returning 404. I've done a local rclone dry run, which shows that the renamed doc is no longer being excluded by `--exclude "assets/**" \`.

PR #964, opened 2026-04-28
Update WebView docs for WPEWebKit

drasticactions:
For https://github.com/AvaloniaUI/Avalonia.Controls.Maui/issues/192

I didn't update the docs for this when I landed it in Avalonia.Controls.Webview. 100% on me. This updates the docs to note that we do support Linux Webviews, and that you should use WPEWebKit first when available.

I tested the install guides on Ubuntu, Fedora, and Arch, and they worked for me.


luke-whos-here (approved):
LGTM, thanks for these!

PR #970, opened 2026-05-05
Add docs for AOT support

luke-whos-here:
- Add new doc for using AOT with XPF specifically.
- Add section to XPF FAQ page on AOT.
- Add crosslink from Avalonia AOT page.
- Update sidebar with link to new page.

### DIFF
--- a/docs/deployment/native-aot.md
+++ b/docs/deployment/native-aot.md
@@ -107,6 +107,10 @@ When using Native AOT with Avalonia, be aware of these limitations:
 
 For platform support, refer to [Platform/architecture restrictions](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/#platformarchitecture-restrictions).
 
+## Avalonia XPF
+
+If you are using [Avalonia XPF](/xpf), Native AOT is also supported. See [XPF: Native AOT](/xpf/deployment/native-aot) for XPF-specific setup and usage.
+
 ## See also
 
 - [Native AOT deployment](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet9plus#platformarchitecture-restrictions): Microsoft documentation on Native AOT.
--- a/docs/deployment/native-aot.md
+++ b/docs/deployment/native-aot.md
@@ -51,48 +51,29 @@ When using Native AOT, XAML is compiled into the application at build time. Ensu
 
 ## Publishing Avalonia Native AOT applications
 
-### Windows
-```bash
-dotnet publish -r win-x64 -c Release
+To publish your app, run `dotnet publish` in the command line:
+
+```
+dotnet publish -r <current platform> -c Release
 ```
 
-### Linux
-```bash
-dotnet publish -r linux-x64 -c Release
-```
+As an example, `dotnet publish -r osx-arm64 -c Release` would publish the app for Apple Silicon devices.
 
-### macOS
-Intel based macOS 
-```bash
-dotnet publish -r osx-x64 -c Release
-```
-
-Apple silicon based macOS 
-```bash
-dotnet publish -r osx-arm64 -c Release
-```
+For more information, please see [Native AOT deployment on the .NET documentation site](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#publish-native-aot-using-the-cli).
 
 :::tip
 You can then use Apple's [lipo tool](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) to combine both Intel and Apple Silicon binaries, enabling you to ship  Universal binaries.
 :::
 
-## Troubleshooting common issues
+## Resolving reflection-related errors
 
-### 1. Reflection-related errors
-For view models or services using reflection:
-```xml
+Add a trimmer root descriptor to your `.csproj` file.
+
+```xml title=".csproj"
 <ItemGroup>
-    <TrimmerRootDescriptor Include="TrimmerRoots.xml" />
+    <ProjectReference Include="..\YourApp\YourApp.csproj" />
+    <TrimmerRootAssembly Include="YourAssembly" />
 </ItemGroup>
-```
-
-Create a `TrimmerRoots.xml`:
-```xml
-<linker>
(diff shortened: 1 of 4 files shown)


xpf/deployment/native-aot.md
```diff
@@ -0,0 +1,91 @@
...
+
+```bash
+dotnet publish -r linux-x64 -c Release
+```
+
+For ARM64 devices:
+
+```bash
+dotnet publish -r linux-arm64 -c Release
+```
```
maxkatz6: I think these commands are less useful to include in the documentation.

.NET AOT only allows publishing for the current platform (i.e., on windows, only for windows). Where only architecture (x64/arm64) is allowed runtime variable.

I would structure it as a single:
```
dotnet publish
```

Where "-c Release" is implied as a default value for publish. And "-r" by default is "current platform".
And link https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#publish-native-aot-using-the-cli for any extra information.

xpf/deployment/native-aot.md
```diff
@@ -0,0 +1,91 @@
...
+```xml
+<linker>
+    <assembly fullname="YourAssembly">
+        <type fullname="YourAssembly.ViewModels*" preserve="all"/>
+        <type fullname="YourAssembly.Views*" preserve="all"/>
+    </assembly>
+</linker>
+```
+
+Adjust the type patterns to match the namespaces your XAML references. Some third-party libraries may require comprehensive entries if they do not ship with a root descriptor.
```
maxkatz6: TrimmerRoots.xml format is not really documented by Microsoft (at least as part of stable documentation, it is documented in the github source code).

More typical approach is using `<TrimmerRootAssembly Include="YourAssembly" />`, documented here https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#csproj-file.

maxkatz6: One difference from the standard SDK, XPF by default includes `TrimmerRootAssembly` for all WPF assemblies and user executable assembly as well (on these assemblies that have `Xpf.Sdk` set). I think it's important to mention, so users won't duplicate it for no reason.

xpf/deployment/native-aot.md
luke-whos-here: I have edited this section as suggested and included the link. Also duplicating the edits to the AOT page for OSS Avalonia.

xpf/deployment/native-aot.md
luke-whos-here: Edited as suggested and added a link to the .NET documentation. Again, I'm duplicating these changes to the OSS Avalonia page on AOT.

docs/deployment/native-aot.md
```diff
@@ -51,50 +51,31 @@ When using Native AOT, XAML is compiled into the application at build time. Ensu
...
-### Linux
-```bash
-dotnet publish -r linux-x64 -c Release
 ```
-
-### macOS
-Intel based macOS 
-```bash
-dotnet publish -r osx-x64 -c Release
+dotnet publish -r <current platform> -c Release
```
maxkatz6: Just `-r <runtime>`, otherwise removing `-r` flag completely would also mean the same as "current platform".

docs/deployment/native-aot.md
```diff
@@ -51,50 +51,31 @@ When using Native AOT, XAML is compiled into the application at build time. Ensu
...
 
-### 1. Reflection-related errors
-For view models or services using reflection:
-```xml
+Add a trimmer root descriptor to your `.csproj` file.
+
+```xml title=".csproj"
 <ItemGroup>
-    <TrimmerRootDescriptor Include="TrimmerRoots.xml" />
+    <ProjectReference Include="..\YourApp\YourApp.csproj" />
```
maxkatz6: I see `<ProjectReference>` is copied from the .NET doc, but it was intended there in combination with TrimmerRootAssembly.

If we add it, it should be:
```
<ProjectReference Include="..\YourApp\YourApp.csproj" />
<TrimmerRootAssembly Include="YourApp" />
```

Or in the same way as in .NET doc - `MyLibrary` everywhere ("YourApp" is less useful in this context, as app itself is already rooted, but libraries might not be).

docs/deployment/native-aot.md
luke-whos-here: Meaning just `dotnet publish -r -c Release`? In that case, should we also remove the Apple example below?

docs/deployment/native-aot.md
maxkatz6, line 29: For normal Avalonia (not XPF), I would also recommend adding `<IsAotCompatible>true</IsAotCompatible>`. Not only that, but also to every project user has, to ensure every project is compatible with AOT (not like `PublishAot`, which only necessary on the main/executable project). See https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8#aot-compatibility-analyzers (could be linked too)

docs/deployment/native-aot.md
maxkatz6: Ah, sorry, github ate my `<`

maxkatz6: Updated my comment

docs/deployment/native-aot.md
luke-whos-here: Apologies - that was me having a brain fart. I had intended to type YourAssembly everywhere, but it somehow came out as YourApp.

docs/deployment/native-aot.md
maxkatz6: The issue here is only in "current platform" wording. "-r <runtime>" flag is only useful, when users want to publish for "non-current platform", because by default it's always "current platform".

docs/deployment/native-aot.md
maxkatz6: For XPF IsAotCompatible _might_ be harmful, because it will produce a lot of false positive warnings/errors. But for normal Avalonia it should be generally safe.

docs/deployment/native-aot.md
luke-whos-here: Got it!

maxkatz6 (approved):

PR #972, opened 2026-05-07
docs: add missing EnumToBoolConverter resource registration to RadioButton example

dhoehna:
## Summary

The **Binding to an enum** section of the RadioButton documentation references \\EnumToBoolConverter\\ via \\{StaticResource EnumToBoolConverter}\\ but never shows how to register it as a resource. This causes confusion for developers who follow the example and get a runtime error because the resource is not found.

## Changes

Added a short explanation and code sample showing the \\<Window.Resources>\\ declaration before the existing RadioButton binding example. Also linked to the [\\EnumToBoolConverter\\ API reference](/api/avalonia/controls/converters/enumtoboolconverter).

### DIFF
--- a/controls/input/buttons/radiobutton.md
+++ b/controls/input/buttons/radiobutton.md
@@ -102,15 +102,19 @@ public partial class OrderViewModel : ObservableObject
 }
 ```
 
-Avalonia includes a built-in [`EnumToBoolConverter`](/api/avalonia/controls/converters/enumtoboolconverter) for this purpose. Before using it in a binding, register it as an application resource in `App.axaml` so it's available throughout your app. You'll also need to declare the converter namespace in the `<Application>` tag:
+Avalonia includes a built-in [`EnumToBoolConverter`](/api/avalonia/controls/converters/enumtoboolconverter) for this purpose. Before using it in a binding, register it as an application resource in `App.axaml` so it's available throughout your app:
 
 ```xml
-<!-- In App.axaml -->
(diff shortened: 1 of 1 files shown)


controls/input/buttons/radiobutton.md
```diff
@@ -102,6 +102,16 @@ public partial class OrderViewModel : ObservableObject
 }
 ```
 
+Avalonia includes a built-in [`EnumToBoolConverter`](/api/avalonia/controls/converters/enumtoboolconverter) for this purpose. Before using it in a binding, register it as a resource in your window or application:
+
+```xml
+<Window.Resources>
+    <EnumToBoolConverter x:Key="EnumToBoolConverter"/>
+</Window.Resources>
+```
```
luke-whos-here, line 118: For the purposes of this example, I would add the converter as an application resource instead, since it's a general-purpose utility that might be used by other controls bound to an enum, depending on the app.

```
<Application.Resources>
    <converters:EnumToBoolConverter x:Key="EnumToBoolConverter"/>
</Application.Resources>
```

I also suggest noting that this should be registered in App.axaml, and that you'd need to declare the namespace in the Application tag - `xmlns:converters="clr-namespace:Avalonia.Controls.Converters;assembly=Avalonia.Controls"`

luke-whos-here (commented):
Checked this and can confirm both the problem and solution are accurate. I have a few minor suggestions for the sample - we can proceed once these are resolved.

luke-whos-here (approved):

PR #974, opened 2026-05-10
Corrected the Avalonia.Templates version to 11.0.2

Abdssamie:
Updated the version of Avalonia.Templates from 11.2.0 to 11.0.2 in the installation instructions. Probably a type

### DIFF
--- a/docs/get-started/install-avalonia.mdx
+++ b/docs/get-started/install-avalonia.mdx
@@ -72,7 +72,7 @@ dotnet new update
 This updates all installed template packages, including `Avalonia.Templates`. To install a specific template version (for example, to match a particular Avalonia release), specify the version explicitly:
 
 ```bash
-dotnet new install Avalonia.Templates::11.2.0
+dotnet new install Avalonia.Templates::11.0.2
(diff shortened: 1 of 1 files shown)


docs/get-started/install-avalonia.mdx
```diff
@@ -72,7 +72,7 @@ dotnet new update
 This updates all installed template packages, including `Avalonia.Templates`. To install a specific template version (for example, to match a particular Avalonia release), specify the version explicitly:
 
 ```bash
-dotnet new install Avalonia.Templates::11.2.0
+dotnet new install Avalonia.Templates::11.0.2
```
luke-whos-here, line 75: Hey, thanks for picking this up! The latest package version of Avalonia.Templates is in fact 12.0.2, so I suggest using that as the example instead of the extremely outdated 11.0.2.

```
dotnet new install Avalonia.Templates@12.0.2
```

luke-whos-here:
This fix was included as part of https://github.com/AvaloniaUI/avalonia-docs/pull/995

PR #982, opened 2026-05-14
NativeWebView content edits

luke-whos-here:
Clarify that GTK is unsupported by NativeWebView on relevant pages.


luke-whos-here:
@maxkatz6 I made some edits to the NativeWebView docs in response to the feedback about no GTK support. Can you have a look at these and confirm if it's clearer?

maxkatz6 (approved):
Thanks!

PR #983, opened 2026-05-15
Change ScatterSeries to ScatterLineSeries in scatterlineseries example

ykarpeev:


luke-whos-here:
This is indeed the intended series for the sample. Thanks for the correction.

luke-whos-here (approved):

PR #984, opened 2026-05-15
980 restore the installation instructions for avalonia pro controls

### LINKED ISSUE #980: Restore the installation instructions for Avalonia Pro controls.
At one point there were installation instructions for Avalonia Accelerate components, e.g. this one for media player:

https://github.com/AvaloniaUI/avalonia-docs/blob/1726f85f159ab1af3303702e71f2dd1ece729995/reference/components/media-player/quickstart.md

However these instructions seem to have been deleted at some point in the preparation of the v12 documentation:

- Delete commit: https://github.com/AvaloniaUI/avalonia-docs/commit/e97b52867b3a9a50461ddd7c4f964818bd7d2351
- Merge PR: https://github.com/AvaloniaUI/avalonia-docs/pull/917

This installation/getting started documentation should be restored as it's impossible to use the Avalonia Pro controls without knowing:

- What package to install
- What themes to add
- Any extra initialization steps

luke-whos-here:
This PR restores "getting started" instructions to the index pages of each Avalonia Pro controls. Instructions have been rewritten to be consistent across each appearance, and to include crosslinks to the Avalonia portal and the "Installing Avalonia Pro" page.

This PR has also been used as an opportunity to restructure the docs into more logical groupings, where relevant. Redirects have been added for pages that were affected by this work.

All instructions have been tested to be working correctly with Avalonia 12.0.x.


controls/data-display/charts/index.md
```diff
@@ -14,9 +14,27 @@ Charts provide a library of data visualization controls and documented compositi
...
+<ItemGroup>
+  <AvaloniaUILicenseKey Include="YOUR_LICENSE_KEY" />
+</ItemGroup>
+```
+
+:::tip
+For multi-project solutions, you can store your licence key in an [environment variable](https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-use-environment-variables-in-a-build) or a [shared props file](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022#directorybuildprops-example) to avoid duplication.
+:::
+
+3. Use the `https://avaloniaui.net/controls/charts` XML namespace in your XAML. All chart types share the same namespace.
```
grokys: Not strictly necessary - one can actually just use the default namespace - and no other controls have a custom XML namespace. Might be worth using the default namespace in the docs and just mentioning that one can use this if they want to? Will let you decide what's clearer.

grokys (approved):
LGTM other than the below comment. I'll let you decide with @jsuarezruiz what the best approach there is.

controls/data-display/charts/index.md
luke-whos-here: Thanks Steven. We will go with your suggestion.

PR #989, opened 2026-05-19
Revise data persistence documentation for clarity

maxkatz6:
Previous documentation with hardcoded paths/conditions per each OS was error prone or just wrong.
I changed `GetAppDataPath` to use `SpecialFolder.LocalApplicationData` always and added several nice common additions (`SpecialFolderOption.Create`, `Directory.CreateDirectory`).


luke-whos-here (approved):
LGTM

PR #996, opened 2026-05-25
Fix links for ToggleSwitch

fralonra:
- Fix links for ToggleSwitch in Home page


luke-whos-here (approved):
Nicely found, thanks for the fix. I've taken this opportunity to also correct the component headers for consistency.

PR #997, opened 2026-05-26
Increased the line-height of search titles

MikeCodesDotNET:
If the search term is too long, it wraps to a new line, which has a line height that is too small, resulting in this horrible-looking result. 

docs avaloniaui net_docs_welcome

I've confirmed it looks correct when searching with the long text. Might be worth a very quick sanity check that this doesn't break something else 👀


luke-whos-here (approved):
Looks okay to me. Thanks for the fix Mike!

PR #1006, opened 2026-06-01
Document vertical separator variant in separator.md

otac0n:
Added documentation for vertical separator variant using Height and Width properties.


luke-whos-here (approved):
Thanks for the contribution. I've checked your example and it works fine.

I've taken this opportunity to also update the page to use the web-based XAML previewer.

PR #1027, opened 2026-06-17
Add missing "Enable XAML Previewer" entry for VS Code

mrlacey:
There were already entries for VS & Rider, but the VSCode entry was missing.


luke-whos-here (approved):
I made a few edits to really emphasize that previewer is now available on every IDE we recommend. Otherwise LGTM.

PR #1028, opened 2026-06-17
Include installation details for VSCode extension

mrlacey:
Previously the install/set up instructions for the VSCode extension were missing. This adds some simple instructions to bring a parallel with the other extensions.

Also, corrects the minimum supported version for the VS extension.


luke-whos-here (approved):
Two very small edits to make things look nice. Otherwise LGTM.

PR #1058, opened 2026-07-07
Add TableView documentation

MrJul:
Adds documentation for the new `TableView` control implemented in https://github.com/AvaloniaUI/Avalonia/pull/21511.

### DIFF
--- a/controls-sidebar.ts
+++ b/controls-sidebar.ts
@@ -30,6 +30,7 @@ const sidebars: SidebarsConfig = {
           label: 'Structured data',
           collapsed: true,
           items:[
+              'data-display/structured-data/tableview',
               'data-display/structured-data/datagrid',
               {
                 type: 'category',
--- a/controls/data-display/structured-data/tableview.md
+++ b/controls/data-display/structured-data/tableview.md
@@ -29,6 +29,7 @@ Bind the `ItemsSource` property to a collection in your view model, then declare
 >
 
 <TabItem value="xaml">
+
 ```xml
 <TableView ItemsSource="{Binding Countries}">
   <TableView.Columns>
@@ -40,9 +41,11 @@ Bind the `ItemsSource` property to a collection in your view model, then declare
   </TableView.Columns>
 </TableView>
 ```
+
 </TabItem>
 
 <TabItem value="viewmodel">
+
 ```csharp
 using System.Collections.ObjectModel;
 
@@ -56,12 +59,15 @@ public class MainWindowViewModel
     ];
 }
 ```
+
 </TabItem>
 
 <TabItem value="item">
+
 ```csharp
 public record Country(string Name, string Region, int Population);
 ```
+
 </TabItem>
 
 </Tabs>
@@ -176,7 +182,6 @@ Dragging a resizer switches that column to a pixel width.
 
 ## Virtualization
 
-Similar to `ListBox`, rows are virtualized and recycled by default.  
 Similar to `ListBox`, rows are virtualized and recycled by default. Cells are also recycled alongside their owning rows.
 
 :::warning
--- a/controls/data-display/structured-data/tableview.md
+++ b/controls/data-display/structured-data/tableview.md
@@ -1,0 +1,168 @@
+---
+id: tableview
+title: TableView
+---
+
+The `TableView` displays a collection of items in configurable columns. It is a read-only, tabular control: it presents data but does not provide in-place cell editing.
+
+`TableView` derives from [`ListBox`](/controls/data-display/collections/listbox), so it reuses the same items source and selection model. Each row is a `TableViewRow`, and each column is defined by a `TableViewColumn`.
+
+:::info
+`TableView` is part of the core **Avalonia.Controls** package, so no additional _NuGet_ package or style include is required. It is available from Avalonia 12.1.
+:::
+
+
+## Basic usage
+
+Bind the `ItemsSource` property to a collection in your view model, then declare one `TableViewColumn` per column inside `TableView.Columns`. Use the column's `Binding` property to pick the value shown in each cell:
+
+```xml title='XAML View'
+<TableView ItemsSource="{Binding Countries}">
+  <TableView.Columns>
+    <TableViewColumn Header="Name" Binding="{Binding Name}" />
+    <TableViewColumn Header="Region" Binding="{Binding Region}" />
+    <TableViewColumn Header="Population"
+                     Binding="{Binding Population, StringFormat=N0}"
+                     HorizontalContentAlignment="Right" />
+  </TableView.Columns>
+</TableView>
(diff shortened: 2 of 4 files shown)


controls-sidebar.ts
maxkatz6, line 34: Would it make sense to update DataGrid docs for the same release?

controls/data-display/structured-data/tableview.md
luke-whos-here:
```suggestion
The `TableView` displays a collection of items in configurable columns. It is a read-only, tabular control: it presents data, but does not provide in-place editing of cell contents.
```

luke-whos-here:
```suggestion
`TableView` is part of the core **Avalonia.Controls** package. No additional NuGet package or style include is required. It is available from Avalonia 12.1.
```

luke-whos-here: Might be nice to put the code blocks in Tabs, instead of a vertical list.

controls-sidebar.ts
MrJul: You mean, to refer to TableView? I can definitely update that.

maxkatz6: Yes

controls/data-display/structured-data/tableview.md
luke-whos-here:
```suggestion
`TableView` derives from [`ListBox`](/controls/data-display/collections/listbox), so it reuses the same `ItemsSource` and `SelectionModel`. Each row is a `TableViewRow`, and each column is a `TableViewColumn`.
```

luke-whos-here, line 88: Should this table include `Rows`, or is the user not expected to need them much?

luke-whos-here:
```suggestion
| `Binding`      | Retrieves the cell value from the row's data item through a binding. Intended for simple property displays, e.g., `Binding="{Binding Name}"`.             |
```

luke-whos-here:
```suggestion
| `CellTemplate` | Builds the cell content using a data template. The whole row data item is passed as the template's data context, so you can bind to any property.   |
```

luke-whos-here:
```suggestion
| `Binding`                    | A binding that reads the cell value from the row data item ([see above](#displaying-cell-values)).                                                                                      |
```

luke-whos-here:
```suggestion
| `CellTemplate`               | A data template for the cell content, receiving the row data item as its data context ([see above](#displaying-cell-values)).                                                          |
```

luke-whos-here:
```suggestion
| `Width`                      | The column width, expressed as a `GridLength` ([see below](#column-width)). The default is `1*`.                                                                              |
```

luke-whos-here:
```suggestion
| `CanUserResize`              | Whether this specific column can be resized. When left at its default (`null`), the value falls back to the `CanUserResizeColumns` property of the `TableView` ([see above](#useful-properties)).                  |
```

luke-whos-here:
```suggestion
By default, users can resize columns by dragging the separator between two column headers. To turn this off for the whole `TableView`, set `CanUserResizeColumns` to `False`.
```

luke-whos-here:
```suggestion
### Resizing columns
```

luke-whos-here:
```suggestion
You can also override the behavior for individual columns with `CanUserResize`. When it is left at its default value of `null`, the column follows the `CanUserResizeColumns` setting of the overall `TableView`. Changing it to `True` or `False` overrides the setting for that column only:
```

luke-whos-here, line 179:
```suggestion
:::info
Dragging a resizer switches that column to a pixel width.
:::
```

luke-whos-here:
```suggestion
Similar to `ListBox`, rows are virtualized and recycled by default. Cells are also recycled alongside their owning rows.
```

controls/index.md
luke-whos-here, line 116: Not a blocker, but it would be nice if we had a new icon for TableView, rather than recycle the Grid icon.

luke-whos-here (commented):
Finished my review. I've left a couple of comments, suggestions and questions.

controls/data-display/structured-data/tableview.md
MrJul: `Rows` does not exist. Like with `ListBox`, the `Items` property can be used to define items manually, but the 99% use case is through binding a list via `ItemsSource`.

controls/index.md
MrJul: I agree, but do we have a tool to make icons in the existing style?

controls-sidebar.ts
MrJul: Updated the DataGrid page.

MrJul:
@luke-whos-here I've applied your suggestions and addressed all comments, except for the icon.

controls/data-display/structured-data/tableview.md
luke-whos-here: Thanks for the explanation. That being the case, should we also remove the mention of `TableViewRow` from line 11, to avoid being misleading?

luke-whos-here:
Just my two cents on the DataGrid doc - is there an argument to actually delete most of the content, leaving only the warning callout box to note that it is deprecated?

controls/data-display/structured-data/tableview.md
MrJul: I think it needs to be mentioned, if only for styling purposes.

MrJul:
> is there an argument to actually delete most of the content, leaving only the warning callout box to note that it is deprecated?

I think that could wait for a little while, but I won't die on that hill. Existing users of `DataGrid` might want to keep it until they choose an alternative solution, and documentation should be available in that case.

controls/index.md
luke-whos-here: I want to say AI, but we should probably ask Dalton to make one.

controls/data-display/structured-data/tableview.md
luke-whos-here: Duplicated sentence here.

luke-whos-here (commented):
All good aside from one weird repeated sentence, probably from me misformatting a suggestion.

controls/data-display/structured-data/tableview.md
MrJul: Yes, all your suggestions were missing a "before", which is a bit odd.
I've fixed the duplicated sentence.

MrJul: Apparently, it works better if I push my changes :)

luke-whos-here (approved):

PR #1075, opened 2026-07-17
Custom control edits

luke-whos-here:
This PR revises the /docs/custom-controls collection with significant updates

- Condenses content into fewer pages, combining related material onto a single page where sensible.
- Adds redirects for retired or renamed pages.
- Verifies and updates all code samples to ensure they work, presenting them with the web XAML previewer where feasible.
- Adds step-by step instructions for tutorial-style pages.
- Rewrites the index page into a true overview, rather than a lengthy all-purpose page.
- Removes sections with excessive theoretical material out of scope for a custom control guide. Add crosslinks instead to relevant conceptual pages.
- Removes or replaces outdated information and screenshots.
- Updates sidebar and links to be consistent with new metadata structure.
- Applies a whole bunch of generalist grammar and style edits.


docs/custom-controls/custom-panel.md
```diff
@@ -77,7 +124,7 @@ public class RadialPanel : Panel
 
         for (int i = 0; i < Children.Count; i++)
         {
-            double angle = 2 * Math.PI * i / Children.Count - Math.PI / 2;
+            double angle = 2 * 3.14159 * i / Children.Count - 3.14159 / 2;
             double x = centerX + radius * Math.Cos(angle) - Children[i].DesiredSize.Width / 2;
```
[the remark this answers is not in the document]
luke-whos-here: `Math.PI` does not compile in the web XAML previewer. Having a running sample was judged to be a higher priority in this case.

docs/custom-controls/custom-itemspanel.md
```diff
@@ -18,51 +32,32 @@ The following controls all support a custom `ItemsPanel`:
...
+
+    ```csharp title="MainWindowViewModel.cs"
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
+
+    namespace MyApp.ViewModels;
+
+    public record Tile (int Size, int TopX, int TopY);
+
+    public partial class MainWindowViewModel : ViewModelBase
```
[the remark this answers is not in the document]
luke-whos-here: The file `MainWindowViewModel.cs` is clearly named in the title field of both steps, and 2 is clearly building on 1 as a foundation. Omitting `Tile` from the second step would create more confusion, not decrease it.

docs/custom-controls/index.md
```diff
@@ -2,241 +2,84 @@
...
+description: Overview of approaches for building custom controls in Avalonia.
 doc-type: overview
 ---
 
 ## Custom controls
 
-A custom control draws itself using the Avalonia graphics system, using basic methods for shapes, lines, fills, text, and many others. You can define your own properties, events and pseudo classes.
-
-Some of the Avalonia built-in controls are like this. For example, the text block control (`TextBlock` class) and the image control (`Image` class).
+Avalonia allows you to customize your own controls, beyond what is available in the [built-in control library](/controls). Define your own properties, events and [pseudoclasses](/docs/styling/pseudoclasses). You can even override visual rendering to draw entirely unique custom controls.
```
grokys: Nit: this feature isn't really about "customising your own controls" as much as "defining your own controls" or "creating your own controls".

docs/custom-controls/index.md
```diff
@@ -2,241 +2,84 @@
...
+This type of control is best for reusable "views" or "pages" that are specific to an application, for example, a "User Details View". It is less suited for general-purpose UI elements.
 
-BasicControls are the foundation of user interfaces - they draw themselves using geometry by overriding the Visual.Render method. Controls such as TextBlock and Image fall into this category.
+To create a custom user control:
 
-:::info
-In WPF/UWP you would inherit from the FrameworkElement class to create a new basic control, but in Avalonia you should inherit from Control. 
-:::
+1. **Define the XAML.** Create a new `UserControl` XAML file. Decide the layout and appearance of the custom control by placing existing controls, setting properties and applying styles.
+2. **Add code-behind.** Optionally, define code-behind logic to handle events, modify behavior, or give the control [styled properties](/docs/custom-controls/defining-properties#registering-a-styled-property).
```
grokys: There's no description of how to give the control styled properties here or in the sample (the linked styled property docs don't cover this specific use-case AFAICS). This should be added, though not necessarily in this PR.

docs/custom-controls/index.md
```diff
@@ -2,241 +2,84 @@
...
 
-:::info
-The naming convention of the property and its backing `AvaloniaProperty` field is important. The name of the field is always the name of the property, with the suffix `Property` appended.
-:::
+To create a custom templated control:
 
-### Using a `StyledProperty` on another class
+1. **Define the control class.** Create a new class that derives from `TemplatedControl`. Decide the behavior, properties and events of the custom control.
+2. **Add a control template.** Create a [control theme](/docs/styling/control-themes) XAML file. Decide the visual appearance of the control.
+3. **Add further styling.** Further customize the look of the control by adjusting the control template or applying additional styles, if desired.
```
grokys, line 52: Do we want to add a link to the "Creating templated controls" page in this section?

docs/custom-controls/index.md
```diff
@@ -2,241 +2,84 @@
...
 
-Sometimes the property you want to add to your control already exists on another control, `Background` being a good example. To register a property defined on another control, you call `StyledProperty.AddOwner`:
+### Basic controls
```
grokys: Not sure I've seen the terminology "basic controls" used before here, and IMO this gives the wrong impression of what they are (to me "basic" suggests that they're "simple"). This type of control is a real niche use-case and should be the least common type of control. In my experience these are more often called "custom-drawn controls" or similar?

sidebars.ts
```diff
@@ -240,21 +240,19 @@ const sidebars: SidebarsConfig = {
...
     {
       type: 'category',
-      label: 'Custom Controls',
+      label: 'Custom controls',
       collapsed: true,
       items: [
         'custom-controls/index',
-        'custom-controls/choosing-a-custom-control-type',
-        'custom-controls/custom-control-class',
+        'custom-controls/custom-drawn-basic-controls',
```
grokys: This should be the last type of custom control described as they're the least common.

I'd say the order should be:

- User controls (simplest)
- Templated controls
- Custom-drawn controls (most niche)

docs/custom-controls/custom-drawn-controls.md
```diff
@@ -0,0 +1,116 @@
...
+- `AffectsMeasure` triggers a new measure pass, which is appropriate when a property changes the control's desired size.
+- `AffectsArrange` triggers a new arrange pass, which is appropriate when a property changes how the control positions its content.
+
+### Manual invalidation
+
+If you need to trigger a repaint in response to something other than a property change (for example, a timer tick), call `InvalidateVisual()` on the control instance. This queues a new render pass for that control.
+
+```csharp
+// Force a repaint from code
+InvalidateVisual();
```
grokys, line 103: Given that the previous section explains invalidating measure and arrange the `InvalidateMeasure` and `InvalidateArrange` methods should probably be mentioned here too?

docs/custom-controls/templated-controls.md
```diff
@@ -27,41 +31,41 @@ public class ToggleLabel : TemplatedControl
...
-            <Border Background="{TemplateBinding Background}" Padding="8">
-                <TextBlock Text="{TemplateBinding LabelText}" />
-            </Border>
-        </ControlTemplate>
-    </Setter>
-</ControlTheme>
+Every templated control must have a default `ControlTheme` that contains its `ControlTemplate`. This control theme is typically placed in a resource dictionary, such as `Themes/Generic.axaml`, and included in your application resources in `App.axaml`.
+
+```xml title="App.axaml"
+<Application.Resources>
```
grokys, line 45: The paragraph says

> This control theme is typically placed in a resource dictionary, such as `Themes/Generic.axaml`, and included in your application resources in `App.axaml`.

But that's not what this example is showing.

docs/custom-controls/templated-controls.md
```diff
@@ -27,41 +31,41 @@ public class ToggleLabel : TemplatedControl
...
+          </ControlTemplate>
+      </Setter>
+  </ControlTheme>
+</Application.Resources>
 ```
 
-Key points:
+Notes:
 
 - `x:Key="{x:Type local:ToggleLabel}"` ensures Avalonia automatically applies this theme to all instances of `ToggleLabel`.
```
grokys: This could have some additional explanation to describe when you'd want to use a different key (doesn't have to be this PR).

docs/custom-controls/templated-controls.md
```diff
@@ -27,41 +31,41 @@ public class ToggleLabel : TemplatedControl
...
-- `TargetType` scopes the theme so that property setters and template bindings resolve against the correct type.
+- `TargetType` scopes the theme, so that property setters and template bindings resolve against the correct type.
 - Inside the `ControlTemplate`, use [`TemplateBinding`](/api/avalonia/data/templatebinding) to bind to properties on the templated control.
 
 ## Template parts
 
-Sometimes a templated control needs to interact with specific elements inside its template. By convention, these elements are named with a `PART_` prefix.
+Sometimes, a templated control needs to interact with specific elements in its template. By convention, these elements are prefixed with `PART_`.
```
grokys:
```suggestion
Sometimes, a templated control needs to interact with specific elements in its template from code. By convention, these elements are prefixed with `PART_`.
```

grokys: The above could perhaps be worded better, but the distinction is important.

docs/custom-controls/templated-controls.md
```diff
@@ -27,41 +31,41 @@ public class ToggleLabel : TemplatedControl
...
 
-Override `OnApplyTemplate` to locate named parts after the template has been applied:
+Override `OnApplyTemplate` to locate named parts after the template has been applied.
```
grokys: We should also touch on the [`[TemplatePart]`](https://github.com/AvaloniaUI/Avalonia/blob/main/src/Avalonia.Base/Controls/Metadata/TemplatePartAttribute.cs) attribute.

docs/custom-controls/templated-controls.md
```diff
@@ -75,11 +79,13 @@ protected override void OnApplyTemplate(TemplateAppliedEventArgs e)
 }
 ```
 
+:::tip
 Because users can replace the control template, always check for `null` when looking up template parts. A custom template may omit parts that the default template provides.
```
grokys: Not correct - template parts can be required or optional, as described by the `IsRequired` property on `TemplatePartAttribute`.

In code, there are two methods: `Find<T>` and `Get<T>`. `Find` is for optional template parts and returns `null` if not found. `Get` is for required parts and throws if not found.

See https://github.com/AvaloniaUI/Avalonia/blob/0b3243e9c074d6d77f8e6fba5b718c0ef89c9d9c/src/Avalonia.Controls/ToggleSwitch.cs#L187 as an example of both in use.

docs/custom-controls/templated-controls.md
```diff
@@ -88,43 +94,45 @@ When you are creating a control template and you want to bind to the templated p
...
 
-    <!-- Instead this syntax must be used in this case -->
+    <!-- Instead, use this syntax for the property path -->
     <TextBlock Name="tb" Text="{Binding Caption.Length, RelativeSource={RelativeSource TemplatedParent}}"/>
     ```
-2.  A `TemplateBinding` only supports `OneWay` mode for performance reasons (this is the [same as WPF](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/templatebinding-markup-extension#remarks)). This means a `TemplateBinding` is actually equivalent to `{Binding RelativeSource={RelativeSource TemplatedParent}, Mode=OneWay}`. If `TwoWay` binding is required in a control template, the full syntax is needed as shown below. Note that `Binding` will also use the default binding mode unlike `TemplateBinding`.
+
+2.  A `TemplateBinding` only supports `OneWay` mode for performance reasons. This is the [same as WPF](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/templatebinding-markup-extension#remarks). A `TemplateBinding` is therefore equivalent to `{Binding RelativeSource={RelativeSource TemplatedParent}, Mode=OneWay}`. If you require `TwoWay` binding in a control template, you must use the full syntax shown below.
```
grokys: Not true for a long time. `TemplateBinding` supports all binding modes except `OneTime` and `OneWayToSource`

https://github.com/AvaloniaUI/Avalonia/blame/0b3243e9c074d6d77f8e6fba5b718c0ef89c9d9c/src/Avalonia.Base/Data/TemplateBinding.cs#L59

docs/custom-controls/templated-controls.md
```diff
@@ -88,43 +94,45 @@ When you are creating a control template and you want to bind to the templated p
...
-<!-- This WON'T work as GeometryDrawing is not a StyledElement. -->
-<GeometryDrawing Brush="{TemplateBinding Foreground}"/>
+3. `TemplateBinding` can only be used on a `StyledElement`.
 
-<!-- Instead this syntax must be used in this case. -->
-<GeometryDrawing Brush="{Binding Foreground, RelativeSource={RelativeSource TemplatedParent}}"/>
-```
+    ```xml
+    <!-- This WON'T work because GeometryDrawing is not a StyledElement -->
+    <GeometryDrawing Brush="{TemplateBinding Foreground}"/>
```
grokys: This is worth checking because I don't think it's true?

docs/custom-controls/templated-controls.md
```diff
@@ -88,43 +94,45 @@ When you are creating a control template and you want to bind to the templated p
...
 
-Set a pseudo-class from your control logic:
+Set a pseudoclass from your control logic.
 
 ```csharp
 PseudoClasses.Set(":active", isActive);
 ```
 
-Then target it in your control theme:
+Then, target it in your control theme.
```
grokys: Worth mentioning the [`PseudoClasses`](https://github.com/AvaloniaUI/Avalonia/blob/0b3243e9c074d6d77f8e6fba5b718c0ef89c9d9c/src/Avalonia.Base/Controls/Metadata/PseudoClassesAttribute.cs) attribute here.

sidebars.ts
```diff
@@ -240,21 +240,19 @@ const sidebars: SidebarsConfig = {
...
         'custom-controls/templated-controls',
         'custom-controls/drawing-custom-controls',
```
grokys: Can this be nested under `custom-drawn-basic-controls`? It only applies to that type of control.

grokys: In fact, I'm not sure it needs to be a separate doc?

docs/custom-controls/defining-properties.md
```diff
@@ -233,20 +247,95 @@ protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs chang
 }
 ```
 
-This approach works for both styled and direct properties.
+## Data validation support
+
+Starting from Avalonia v12, data validation is handled automatically when `enableDataValidation` is set to `true`. The framework detects validation errors from `INotifyDataErrorInfo` and displays them without requiring additional overrides in your control.
```
grokys: Is "Starting from Avalonia v12" correct? I could have sworn that this has worked for many versions before v12?

docs/custom-controls/defining-properties.md
```diff
@@ -233,20 +247,95 @@ protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs chang
...
+This is an example of a custom `TextBox.TextProperty` that reuses `TextBlock.TextProperty`, but adds data validation.
+
+```csharp
+public static readonly DirectProperty<TextBox, string?> TextProperty =
+    TextBlock.TextProperty.AddOwnerWithDataValidation<TextBox>(
+        o => o.Text,
+        (o, v) => o.Text = v,
+        defaultBindingMode: BindingMode.TwoWay,
+        enableDataValidation: true);
+```
```
grokys, line 284: TODO: I think we need documentation on how to actually surface data validation errors in custom controls using [`DataValidationErrors`](https://github.com/AvaloniaUI/Avalonia/blob/main/src/Avalonia.Controls/DataValidationErrors.cs).

docs/custom-controls/defining-properties.md
```diff
@@ -233,20 +247,95 @@ protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs chang
...
+
+## Attached properties
+
+An attached property lives in its own container class and is configured on compatible controls in XAML. This allows you to have additional properties that are not part of your custom control's own control class.
```
grokys: I feel like this description could be more useful, it doesn't really explain why attached properties are useful IMO.

docs/custom-controls/defining-properties.md
```diff
@@ -1,46 +1,48 @@
...
-### Registering a new styled property
+### Registering a styled property
+
+To register a styled property:
+
+1. Add a `static readonly` field of type `StyledProperty<T>`.
+2. Use the method `AvaloniaProperty.Register` to register.
+3. Provide a CLR getter and setter that call `GetValue` and `SetValue` respectively.
 
 The following example registers a `CornerRadius` styled property with a default value of `0.0`:
```
grokys, line 48: Worth adding a warning that the CLR property getters/setters should not do _anything_ but call `GetValue` and `SetValue` because not all property changes go via the CLR property.

docs/custom-controls/defining-events.md
```diff
@@ -1,28 +1,22 @@
...
-title: Defining events
-description: Define and raise routed events on custom Avalonia controls with tunneling and bubbling strategies.
+title: Defining events for custom controls
+sidebar_label: Defining events
+description: Define and raise routed events on custom Avalonia controls.
 doc-type: how-to
 ---
 
-Events in Avalonia allow your custom controls to communicate and notify users of specific actions or occurrences. By defining events, you provide a way for users of your controls to respond and react to these events within their applications. This document will guide you through the process of defining events for your custom controls.
+## Routed events: Summary
```
grokys: Nit: I haven't seen this "Topic: Section" header format elsewhere? Every other page just places the introduction under the title?

docs/custom-controls/control-trees.md
```diff
@@ -1,82 +1,27 @@
...
 doc-type: explanation
 ---
 
-import ControlTreesLogicalScreenshot from '/img/concepts/ui-concepts/controls/control-trees-logical.png';
-import ControlTreesVisualScreenshot from '/img/concepts/ui-concepts/controls/control-trees-visual.png';
-import ControlTreesEventScreenshot from '/img/concepts/ui-concepts/controls/control-trees-events.png';
+## Control trees: Summary
 
-Avalonia creates control trees from the XAML files in an application so that it can render the UI presentation and manage the application functionality.
+Avalonia organizes controls into two parallel tree structures: the **logical tree** and the **visual tree**. The logical tree represents the hierarchy of the application's controls, whereas the visual tree represents all visual elements that are being rendered.
```
grokys: Hmm, they're not really _parallel_. Instead the logical tree is a subset of the visual tree (kinda... mostly - it's complicated).

docs/custom-controls/control-trees.md
```diff
@@ -1,82 +1,27 @@
...
-    predicate: p => p.IsEnabled);
-var visibleTextBox = myControl.FindDescendantOfType<TextBox>(
-    predicate: tb => tb.IsVisible);
-
-// Get all logical children
-var children = myControl.GetLogicalChildren();
-```
+For more information on control trees, see [Visual and logical trees](/docs/fundamentals/visual-and-logical-trees).
 
 ## Tree traversal in custom controls
```
grokys: This section doesn't describe traversal at all? It's describing attachment and detachment events.

docs/custom-controls/control-trees.md
```diff
@@ -1,82 +1,27 @@
 ---
 id: control-trees
-title: Control trees
-description: Understand how Avalonia organizes controls into logical and visual trees for layout and rendering.
+title: Traversing control trees with custom controls
```
grokys: This document doesn't describe tree traversal?

docs/custom-controls/control-trees.md
```diff
@@ -92,13 +37,15 @@ protected override void OnDetachedFromVisualTree(VisualTreeAttachmentEventArgs e
 }
 ```
 
-### Where to mutate `LogicalChildren`
+## Where to mutate `LogicalChildren`
 
 When a custom container needs to add or remove logical children programmatically (for example, inserting separators between items), do this inside a structured lifecycle hook, e.g., `OnApplyTemplate`, rather than a property-change callback like `DataContextChanged` or `OnPropertyChanged`. Mutating `LogicalChildren` while the logical tree is being walked can result in binding errors.
```
grokys, line 39: Ohhh god. This subject should really be a whole document... There should really be _very_ little reason to do any of this, but when you _do_ find yourself in a position where you need to do this, you're in deep water. Consider this a TODO ;)

docs/custom-controls/custom-flyout.md
```diff
@@ -1,69 +1,121 @@
...
+```xml title="MainWindow.axaml"
+<Window xmlns="https://github.com/avaloniaui"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:local="using:CustomFlyoutDemo"
+        x:Class="CustomFlyoutDemo.MainWindow">
+    <Button Content="Show image"
+            HorizontalAlignment="Center"
+            VerticalAlignment="Center">
+        <Button.Flyout>
+            <local:MyImageFlyout Image="{StaticResource MyPicture}" />
```
grokys, line 104: When will this flyout be shown? The document covers manual hiding and showing but doesn't mention what interactions will cause the flyout to be shown.

docs/custom-controls/custom-panel.md
```diff
@@ -1,25 +1,47 @@
...
+Every panel must participate in both passes.
 
-This simple `PlotPanel` positions all child elements at a hard-coded offset of (50, 50).
+For more information on Avalonia's layout system, see [The layout system](/docs/layout/#the-layout-system).
+
+## Creating a custom `PlotPanel` with a fixed offset
+
+This simple example highlights how to override `MeasureOverride` and `ArrangeOverride` when creating a custom panel. The custom `PlotPanel` positions child elements at a hard-coded offset of (50, 50).
+
+<XamlPreview>
```
grokys, line 26: Niiice! Love that there's runnable code in here! We need more of this!

docs/custom-controls/custom-panel.md
```diff
@@ -87,21 +134,13 @@ public class RadialPanel : Panel
...
-    <Button Content="2" />
-    <Button Content="3" />
-    <Button Content="4" />
-    <Button Content="5" />
-</local:RadialPanel>
-```
+## Adding an attached property
 
-## Using attached properties
+Add a `Slot` property to `RadialPanel`, so that children can specify their position in the circle. An attached property is chosen to allow data to be specified on a per-child basis.
```
grokys, line 141: Now _this_ is an excellent example of _why_ you'd want to use an attached property!

docs/custom-controls/custom-panel.md
```diff
@@ -111,19 +150,16 @@ public static int GetSlot(Control element) => element.GetValue(SlotProperty);
...
 ```
 
-The panel can then read `GetSlot(child)` inside `ArrangeOverride` to determine where each child should be placed.
-
-For a full guide on defining attached properties, see [Attached Properties](/docs/custom-controls/attached-properties).
-
 ## Tips
 
 - Always call `Measure` on every child in `MeasureOverride`. Children that are not measured will not render correctly.
 - Always call `Arrange` on every child in `ArrangeOverride`. Children that are not arranged will not appear.
```
grokys, line 156: Not strictly true, but probably gets the idea across better than the reality.

docs/custom-controls/custom-panel.md
```diff
@@ -111,19 +150,16 @@ public static int GetSlot(Control element) => element.GetValue(SlotProperty);
...
-- Use `AffectsMeasure` or `AffectsArrange` when registering styled properties that influence layout. This ensures the panel re-layouts when those properties change.
+- Use `AffectsMeasure` or `AffectsArrange` when registering styled properties that influence layout. This ensures the panel adjusts its layout when those properties change.
```
grokys: Link to docs for `AffectsMeasure` and `AffectsArrange`?

docs/custom-controls/custom-itemspanel.md
```diff
@@ -1,15 +1,29 @@
 ---
 id: custom-itemspanel
 title: Custom ItemsPanel
-description: Replace the default items panel in an ItemsControl with a custom panel like Canvas.
+description: Replace the default items panel in an ItemsControl with a custom panel to achieve unique layouts not provided by the built-in controls.
```
grokys: To me, this document doesn't really fit here. It doesn't describe creating a custom control at all - there _is_ a case where one wants to write their own `Panel` class to use in an `ItemsPanel` but this document doesn't describe that.

docs/custom-controls/custom-control-library.md
```diff
@@ -1,30 +1,34 @@
...
+import NewClassLibraryRider from '/img/custom-controls/new-class-library-rider.png';
+import NewClassLibraryVSCode from '/img/custom-controls/new-class-library-vscode.png';
+import InstallAvaloniaInClassLibraryVS from '/img/custom-controls/install-avalonia-in-class-library-vs.png';
+import InstallAvaloniaInClassLibraryRider from '/img/custom-controls/install-avalonia-in-class-library-rider.png';
+import InstallAvaloniaInClassLibraryVSCode from '/img/custom-controls/install-avalonia-in-class-library-vscode.png';
+import CustomControlSolution from '/img/custom-controls/custom-control-solution.png';
+import CustomControlPreview from '/img/custom-controls/custom-control-preview.png';
 
-This guide shows you how to create a custom control library and reference it for use in an _Avalonia UI_ app.
+This guide shows you how to create a custom control library and reference it for use in an Avalonia app.
```
grokys: This document doesn't describe the case of using a templated control from a library, where one has to `<StyleInclude>` the theme.

docs/custom-controls/custom-control-library.md
```diff
@@ -170,27 +151,34 @@ You can now make a namespace declaration in .axaml files of your Avalonia projec
 
 <Image light={CustomControlPreview} alt="A screenshot of an IDE, displaying XAML code in one window and a preview of a user interface in another." position="center" maxWidth={400} cornerRadius="true"/>
 
-## XML Namespace Definitions
+## XML namespace definitions
 
-When referencing a control library in a .axaml file, you can use the URL identification format. For example:
+When referencing a control library in a `.axaml` file, you can use the URL identification format. For example:
```
grokys, line 276: Maybe contrast this with the `using:` or `clr-namespace:` URL formats?

docs/custom-controls/custom-control-library.md
grokys: Not a huge fan of this document in general, something seems off but I can't quite put my finger on it. Something about the way it jumps from basics (creating a project - we're at the end of a section that describes many advanced features, why do we need to go back to basics _here_?) to creating custom XML URLs without describing what they are or why you'd want them, or what the alternatives are.

It's half "getting started" and half "advanced concepts" with nothing inbetween.

grokys (changes requested):
Much improved - I particularly love the panels section with runnable code. Given that these docs are about creating controls, which is inherently a _visible_ endeavour I'd like to see more runnable code samples where it makes sense. Having said that, that can be done later - I've mostly just pointed out inaccuracies and missing info in this review.

docs/custom-controls/custom-panel.md
grokys: Hmm looking back at "defining properties" that doc doesn't seem to mention this stuff. Should probably be there,

docs/custom-controls/control-trees.md
luke-whos-here: Will change the title to "Control trees and custom controls".

docs/custom-controls/control-trees.md
luke-whos-here: Yes, that was perhaps poorly expressed. I'll change it to "related" tree structures.

docs/custom-controls/control-trees.md
luke-whos-here: Will change the section heading to "Attachment/detachment events for custom controls".

docs/custom-controls/control-trees.md
luke-whos-here: I agree. I will make a note to revisit the topic as a separate task.

docs/custom-controls/custom-drawn-controls.md
luke-whos-here: Added to the codeblock, and updated the preceding paragraph to describe invalidation more generally.

docs/custom-controls/custom-flyout.md
luke-whos-here: Added an extra bit to the intro to briefly explain the default behaviour of `PopupFlyoutBase`.

docs/custom-controls/custom-panel.md
luke-whos-here: Added.

docs/custom-controls/defining-events.md
luke-whos-here: My thinking at the time I wrote this was I wanted to be clear that this was just a brief summary and not the full conceptual article. But reading this again, I think you're right and it doesn't really need a subtitle at all.

docs/custom-controls/defining-properties.md
luke-whos-here: Added.

docs/custom-controls/defining-properties.md
luke-whos-here: I got this from the [v12 breaking changes doc](https://docs.avaloniaui.net/docs/avalonia12-breaking-changes#data-validation-enabled-by-default-in-custom-controls), although I may be expressing it inelegantly.

> Before v12, Avalonia properties with `enableDataValidation: true` required overriding the `UpdateDataValidation` method for data validation errors to be reported for that property. Now it happens automatically.

docs/custom-controls/defining-properties.md
luke-whos-here: I'll elaborate this by using the custom radial panel as an example, with a crosslink to the relevant page.

docs/custom-controls/index.md
luke-whos-here: Will go with "create".

docs/custom-controls/index.md
luke-whos-here: Added.

docs/custom-controls/index.md
luke-whos-here: I do actually like "custom-drawn" more, so let's go with that.

sidebars.ts
luke-whos-here: Will do.

sidebars.ts
luke-whos-here: Reading the pages again, it's pretty clear they should really be combined into one. Unsure what I was thinking at the time.

docs/custom-controls/templated-controls.md
luke-whos-here: Corrected this to only say it goes in the application resources.

docs/custom-controls/defining-properties.md
luke-whos-here: Expanded this section to elaborate on how this is done.

docs/custom-controls/custom-control-library.md
luke-whos-here: Revised this section with a summary of the formats, with a crosslink to a more detailed page.

docs/custom-controls/templated-controls.md
luke-whos-here: Added an extra bullet under this to explain how to use alternative keys, with a crosslink to control-themes.

docs/custom-controls/templated-controls.md
luke-whos-here: Expanded this section to describe template parts in more detail, with a link to the relevant API doc.

docs/custom-controls/templated-controls.md
luke-whos-here: Corrected this tip as well as the preceding sample to explain the use of `Get` and `Find`.

docs/custom-controls/templated-controls.md
luke-whos-here: Corrected this section, along with other mentions of binding mode on a few other pages.

docs/custom-controls/templated-controls.md
luke-whos-here: Tested this and found it's true. TemplateBinding on GeometryDrawing doesn't prevent compile, but renders the drawn shape without a fill.

docs/custom-controls/templated-controls.md
luke-whos-here: Expanded this section with a more detailed explanation of pseudoclasses and a cross-link to the API reference.

controls/primitives/usercontrol.md
```diff
@@ -5,46 +5,214 @@ description: A base class for creating reusable composite controls with a predef
...
+        </StackPanel>
+    </StackPanel>
+</UserControl>
+```
+
+</XamlPreview>
+
+### Reusing the confirmation view
+
+To reuse the confirmation view in your app, set an `x:Class` in the opening tag of `<UserControl>`. You can then reference its namespace in a `Window` or any other container to bring up another instance of the same user control.
```
grokys: Hmm, although this gives useful information - it isn't strictly correct. What the `x:Class` directive is actually doing is associating a XAML file with a class in code, allowing the code to load the XAML by calling `InitializeComponent`.

So while this _does_ have the side-effect that you can then include the `UserControl` in another view, it feels rather hand-wavy. 

I think instead the `x:Class` stuff should be folded into the Adding code-behind section (as you can't add code-behind until they XAML is associated with the class) and the "reusing" section should be moved to the end, and should simply point out that you can use the class name with an xmlns to include the view in other views.

controls/primitives/usercontrol.md
```diff
@@ -5,46 +5,214 @@ description: A base class for creating reusable composite controls with a predef
...
+    }
+}
+
+```
+
+</XamlPreview>
+
+### Adding a styled property
+
+This example creates a styled property named `Title`, which displays a title at the top of `ConfirmationView`. Using a styled property means the title is variable and can be set to something different per instance of the `ConfirmationView`.
```
grokys: You don't strictly need a styled property to allow it to be set per-instance. The main reason to use a styled properties allows is to support _binding_ (though it also means it can be styled) (and if you don't want it to be styled. a direct property would also allow styling). Maybe to keep it simple at this point though, just point out that it allows binding ;)

controls/primitives/usercontrol.md
```diff
@@ -5,46 +5,214 @@ description: A base class for creating reusable composite controls with a predef
...
+
+:::warning
+You must explicitly name the root element in the `<UserControl>` tag, and then call it wherever you wish the styled property to apply. Doing so ensures the template can reference the control's properties. `root` is highlighted in the sample below to demonstrate how this is done.
```
grokys: This comment isn't clear to me, I think it could be phrased better, e.g. (wording off the top of my head, please tweak):

> The styled property is declared on the root `UserControl` element, so to use it in a binding you must reference that element.  `root` is highlighted in `ConfirmationView.axaml` below to demonstrate how this is done.

controls/primitives/usercontrol.md
```diff
@@ -5,46 +5,214 @@ description: A base class for creating reusable composite controls with a predef
...
+    }
+}
+```
+
+</TabItem>
 
-`UserControl` is the standard approach for creating views in MVVM applications. Each view in your application is typically a `UserControl` subclass paired with a corresponding view model. If you need a control that supports custom templates and theme styling, consider creating a [`TemplatedControl`](/docs/custom-controls/templated-controls) instead.
+</Tabs>
 
 ## See also
```
grokys, line 222: We probably need a section covering binding to a view model/data context here at some point, or at least linking to such a document. I don't want to hold up this PR and longer though so that should come later.

sidebars.ts
```diff
@@ -240,20 +240,21 @@ const sidebars: SidebarsConfig = {
...
+      label: 'Custom controls',
       collapsed: true,
       items: [
         'custom-controls/index',
-        'custom-controls/choosing-a-custom-control-type',
-        'custom-controls/custom-control-class',
+        {
+          type: 'link',
+          label: 'User controls',
+          href: '/controls/primitives/usercontrol',
```
grokys: This may just be me, but clicking on this header in the sidebar causing a jump to a completely different part of the docs is a little surprising... Not sure how to better handle it though?

docs/custom-controls/templated-controls.md
```diff
@@ -27,59 +31,104 @@ public class ToggleLabel : TemplatedControl
...
-<ControlTheme x:Key="{x:Type local:ToggleLabel}" TargetType="local:ToggleLabel">
-    <Setter Property="Template">
-        <ControlTemplate>
-            <Border Background="{TemplateBinding Background}" Padding="8">
-                <TextBlock Text="{TemplateBinding LabelText}" />
-            </Border>
-        </ControlTemplate>
-    </Setter>
-</ControlTheme>
+Every templated control must have a default `ControlTheme` that contains its `ControlTemplate`. This control theme is typically placed in your application resources in `App.axaml`.
```
grokys: Not quite true - the control theme is usually placed in a separate XAML file which is included here. Including it in `App.axaml` is fine for a tutorial but it's not "typical".

docs/custom-controls/templated-controls.md
```diff
@@ -27,59 +31,104 @@ public class ToggleLabel : TemplatedControl
...
 - Inside the `ControlTemplate`, use [`TemplateBinding`](/api/avalonia/data/templatebinding) to bind to properties on the templated control.
 
 ## Template parts
 
-Sometimes a templated control needs to interact with specific elements inside its template. By convention, these elements are named with a `PART_` prefix.
+Sometimes, a templated control needs to interact with specific elements in its template from code. You can specify these elements by defining them as [`TemplatePart`](/api/avalonia/controls/metadata/templatepartattribute). By convention, template parts are named with the `PART_` prefix.
+
+### Declaring parts
+
+Apply [`TemplatePart`](/api/avalonia/controls/metadata/templatepartattribute) to the control class. Create one `TemplatePart` for each element that should be recognized as an individual part.
```
grokys:
```suggestion
Apply [`TemplatePart`](/api/avalonia/controls/metadata/templatepartattribute) attribute to the control class. Create one `TemplatePart` for each element that should be recognized as an individual part.
```

grokys (changes requested):
Nearly there! A few more comments.

controls/primitives/usercontrol.md
luke-whos-here: I will add a link to the Data context doc.

sidebars.ts
luke-whos-here: I can give it the full https:// path, which will open the link in a new tab and also display the "leaving this section" icon.

grokys (approved):

PR #1085, opened 2026-07-27
Fix links in textblock.md and update example

mrlacey:
Updated the property table so that the links were for the types, not the properties, as the links go to type descriptions

Adjusted one of the XAML preview examples for better clarity. I simplified the XAML to make it easier to see and focus on what was different about the elements that are part of the example.


luke-whos-here (approved):
All good!

PR #1102, opened 2026-08-07
Adjust block heights in gradient examples

maxkatz6:
Fixes https://github.com/AvaloniaUI/avalonia-docs/discussions/1100#discussioncomment-17932432

From
image

To

image


luke-whos-here (approved):
LGTM

PR #1107, opened 2026-08-11
XPF 1.6.6 release notes

miloush:


xpf/version-info/release-notes.md
```diff
@@ -3,6 +3,19 @@ id: release-notes
 title: Release notes
 ---
 
+## XPF 1.6.6 (2026-08-11)
+
+* Avalonia version updated from 11.3.18 to 11.3.20
+* Added support for MahApps.Metro
```
luke-whos-here: Re: Emmanuel's comment on Slack, should we change this to "partial support"?

xpf/version-info/release-notes.md
```diff
@@ -3,6 +3,19 @@ id: release-notes
...
+* Fixed fail fast when `FontManagerOptions.DefaultFamilyName` does not point to a system font
+* Fixed `OutOfRangeException` on text input (macOS 27)
```
luke-whos-here: Is it worth grouping all the platform-specific updates as sub-bullets under a single top-level bullet?

luke-whos-here (commented):
Just two questions to confirm with you before we proceed with these.

PR #1124, opened 2026-08-21
Fix docs note on relative points in gradients start and end

emmauss:
Fix a comment in https://docs.avaloniaui.net/docs/graphics-animation/gradients#startpoint-and-endpoint where it says that both percentages and decimals are relative and equivalent, which is incorrect. In Avalonia, decimal values are parsed as absolute units.

### DIFF
--- a/docs/graphics-animation/gradients.md
+++ b/docs/graphics-animation/gradients.md
@@ -30,7 +30,7 @@ All gradient brushes share the `GradientStops` collection and `SpreadMethod` pro
 
 ### `StartPoint` and `EndPoint`
 
-These two properties control the direction of the gradient. You can specify values as percentages (for example, `"0%,50%"`) or as absolute points in the bounding box (for example, `"0,0.5"`).
+These two properties control the direction of the gradient. You can specify values as percentages (for example, `"0%,50%"`) or as absolute points in the bounding box (for example, `"10,15"`).
 
 Common direction patterns:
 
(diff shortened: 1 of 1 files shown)


docs/graphics-animation/gradients.md
```diff
@@ -30,7 +30,7 @@ All gradient brushes share the `GradientStops` collection and `SpreadMethod` pro
 
 ### `StartPoint` and `EndPoint`
 
-These two properties control the direction of the gradient. You can specify values as percentages (for example, `"0%,50%"`) or as decimals relative to the bounding box (for example, `"0,0.5"`). Both forms are equivalent.
+These two properties control the direction of the gradient. You can specify values as percentages (for example, `"0%,50%"`) or as absolute points in the bounding box (for example, `"0,0.5"`).
```
maxkatz6:
```suggestion
These two properties control the direction of the gradient. You can specify values as percentages (for example, `"0%, 50%"`) or as absolute points in the bounding box (for example, `"10, 15"`).
```

I think using "10, 15" as an example is better for absolute units. "0, 0.5" absolute pixels is a bit unpractical typically.

luke-whos-here: I agree with using "10, 15" for the example. It seems clearer if the point is that they're not equivalent anyway.

emmauss: Fixed

luke-whos-here (approved):

PR #1125, opened 2026-08-22
Document method overloading rules and improve docs on binding commands in general

### LINKED ISSUE #1087: Document binding-to-method overload resolution
See https://github.com/AvaloniaUI/Avalonia/pull/21867 for details.
Don't do anything until it's merged, as the exact algorithm may change depending on the review.

(This issue is mostly a self-reminder.)

luke-whos-here:
- Add new content to document method overloading selection rules.
- Add new content to document binding commands to methods in general (which were previously under-documented).
- Remove redundant content on creating commands (which were repeated multiple times on separate pages).
- Correct outdated sections to present up-to-date information on compiled bindings being on by default.
- Add crosslinks to relevant pages.
- General style edits.

### DIFF
--- a/docs/data-binding/binding-debugging.md
+++ b/docs/data-binding/binding-debugging.md
@@ -5,11 +5,9 @@ description: Debug data binding errors in Avalonia using trace-level logging and
 doc-type: how-to
 ---
 
-When a data binding does not produce the expected result, Avalonia provides tools and techniques to diagnose the problem.
-
 ## Binding error logging
 
-Avalonia logs binding errors to the trace output. In a debug build, these messages appear in the IDE Output window or the console. A typical binding error looks like:
+Avalonia logs binding errors to the trace output. In a debug build, these messages appear in the IDE Output window or the console. Here is an example of a typical binding error:
 
 ```text
 [Binding] Error in binding to 'Avalonia.Controls.TextBlock'.'Text': 'Could not find a matching property accessor for 'UserNam' on 'MyApp.ViewModels.MainViewModel'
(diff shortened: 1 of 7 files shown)


docs/data-binding/binding-debugging.md
```diff
@@ -5,118 +5,116 @@ description: Debug data binding errors in Avalonia using trace-level logging and
...
+- Runtime exception when the command runs.
 
-**Fix**: Return an actual value or `null` (which triggers `TargetNullValue` instead).
+**Causes:**
+- Two or more overloads take one parameter and none takes `object`, so the choice is ambiguous.
+- Every overload takes two or more parameters, which method binding does not support.
+- The command parameter does not match the method parameter type, when using a compiled binding. (Compiled bindings do not convert `CommandParameter`.)
 
-## Using Avalonia DevTools
+**Fix:** Give the method a single `object` parameter, or remove the competing overloads. Make sure `CommandParameter` supplies the exact type. See [Binding directly to a method](/docs/data-binding/binding-to-commands#binding-directly-to-a-method) for the full rules.
```
MrJul: I find "Give the method a single `object` parameter" to be unclear: which overload is the user supposed to modify? The suggested fix should be to remove the competing overloads first, or to add a new one taking an `object` parameter second.

docs/data-binding/compiled-bindings.md
```diff
@@ -5,49 +5,42 @@ description: Use compiled bindings for compile-time validation and improved perf
...
-* If you use compiled bindings and the property you bind to is not found, you get a compile-time error. This gives you a much better debugging experience.
-* Reflection is known to be slow. Using compiled bindings can therefore improve the performance of your application.
+* If the property you bind to is not found, you get a compile-time error to aid debugging.
+* Reflection is known to be slow. Compiled bindings can improve the performance of your application.
 
-## Enable and disable compiled bindings
+## Enabling and disabling compiled bindings
 
-:::info
+Since version 12, Avalonia templates set `<AvaloniaUseCompiledBindingsByDefault>` to `true` by default. This means you only need to provide `x:DataType` for the objects you want to bind to, and do not need to set `x:CompileBindings="[True|False]"` on controls and windows.
```
MrJul: That isn't true anymore.

For history:
 - In v11, the default was `false`, but templates were adding and setting the flag to `true`.
 - In v12, the default is now `true`, and templates don't add the flag anymore.

So users won't find `AvaloniaUseCompiledBindingsByDefault` in their csproj when using the templates (see https://github.com/AvaloniaUI/Avalonia.Templates/pull/341/). They can still add the flag and set it to false if needed, but the documentation should explicitly discourage that.

MrJul (changes requested):

docs/data-binding/binding-debugging.md
```diff
@@ -96,7 +96,11 @@ public static AppBuilder BuildAvaloniaApp()
 - Every overload takes two or more parameters, which method binding does not support.
 - The command parameter does not match the method parameter type, when using a compiled binding. (Compiled bindings do not convert `CommandParameter`.)
 
-**Fix:** Give the method a single `object` parameter, or remove the competing overloads. Make sure `CommandParameter` supplies the exact type. See [Binding directly to a method](/docs/data-binding/binding-to-commands#binding-directly-to-a-method) for the full rules.
+**Fix:**
+1. Remove competing overloads.
+2. Add a new overload taking a single `object` parameter (which is chosen with higher priority).
```
MrJul: It looks like the two steps are part of a single solution, but they're in fact different fixes. Sorry if that wasn't clear from my previous message.

luke-whos-here: Apologies, I thought the first/second order was important to note. Combined these into one paragraph now.

MrJul (approved):

PR #1144, opened 2026-09-09
1142 itemscontrol virtualization

### LINKED ISSUE #1142: ItemsControl virtualization
[ItemsControl](https://github.com/AvaloniaUI/avalonia-docs/blob/main/controls/data-display/collections/itemscontrol.md) doc suggest using ItemsRepeater for virtualization, which is "retired" and not even in the default package anymore (also not stated).

Instead, it should probably showcase usage with VirtualizingStackPanel.

luke-whos-here:
This PR removes most mentions of ItemsRepeater from the docs. Samples featuring ItemsRepeater are replaced with equivalents using other controls where appropriate. They are removed entirely in some cases where retaining the sample is deemed to be of little value.

A single note acknowledging the retirement of ItemsRepeater is retained on the ItemsControl page.

The main ItemsRepeater reference page is removed and will redirect to the ItemsControl page.


controls/data-display/collections/itemscontrol.md
```diff
@@ -18,36 +16,21 @@ You will probably use these properties most often:
...
-
-- Because `ItemsControl` has no built-in scrollbar, content that overflows the available height is clipped. Wrap the control in a `ScrollViewer` if you need scrolling:
-
-```xml
-<ScrollViewer>
-  <ItemsControl ItemsSource="{Binding MyItems}" />
-</ScrollViewer>
-```
+- **Use `ObservableCollection<T>`** for your `ItemsSource` if you need the UI to update automatically when you add or remove items at runtime. A plain `List<T>` will not update the control when changed.
+- **`ItemsControl` does not virtualize** its children. If you are working with a large number of items, consider consider wrapping your `ItemsControl` in a [`ScrollViewer`](/controls/layout/containers/scrollviewer) so that it scrolls. Or use a [`ListBox`](/controls/data-display/collections/listbox), which virtualizes by default.
```
MrJul: > **`ItemsControl` does not virtualize** its children. If you are working with a large number of items, consider consider wrapping your `ItemsControl` in a [`ScrollViewer`](/controls/layout/containers/scrollviewer) so that it scrolls.

This isn't good advice, adding a `ScrollViewer` won't add virtualization: all items will still be laid out, causing slowness.  Keep only the `ListBox` part.

controls/data-display/collections/itemsrepeater.md
MrJul, line 1: Is it worth keeping this page with minimal content ("don't use this") and a link to `ItemsControl` so it appears in searches?

controls/layout/panels/stackpanel.md
```diff
@@ -123,7 +123,7 @@ You can create a horizontal button bar by setting `Orientation` to `Horizontal`
 ## Practical notes
 
 - **Sizing behavior**: A `StackPanel` does not constrain children in the stacking direction, so each child receives as much space as it requests. If you need children to share space proportionally, consider using a `Grid` instead.
-- **Performance**: For lists with many items, prefer `ItemsRepeater` or `ListBox` with virtualization over placing large numbers of controls in a `StackPanel`.
+- **Performance**: For lists with many items, use `VirtualizingStackPanel` or `ListBox` instead of placing many controls in a `StackPanel`.
```
MrJul: It can be mentioned that `ListBox` uses a `VirtualizingStackPanel` internally, but we can't advise users to use `VirtualizingStackPanel` directly: this panel only works when an outer control (the `ListBox`) can drive the items to be realized

Otherwise, users might replace `StackPanel` with `VirtualizingStackPanel` and be confused when nothing displays.

docs/how-to/itemscontrol-how-to.md
```diff
@@ -1,25 +1,24 @@
...
+Use `ItemsControl` when you need to display a small-to-medium collection without selection behavior. Use `ListBox` when you need a virtualized, selectable collection.
 
 | Control | Selection | Virtualization | Best For |
 |---|---|---|---|
 | `ListBox` | Built-in | Yes | Selectable lists |
-| `ItemsControl` | None | No (by default) | Small collections, custom layouts |
-| `ItemsRepeater` | None | Yes | Large collections, custom layouts, performance |
-
-Use `ItemsControl` when you need to display a collection without selection behavior. Use `ItemsRepeater` when you also need virtualization for large data sets.
+| `ItemsControl` | None | No | Custom layouts |
```
MrJul: Note: removing `No (by default)` might lead people to think it isn't possible at all. An `ItemsControl` + `VirtualizingStackPanel` as its `ItemsPanel` + a template containing a `ScrollViewer` is equivalent to a `ListBox` without selection.

I'm not asking that we add that here, but it _might_ be worth mentioning in an advanced topic page, e.g., the `Custom Panel` one. (Can be done in another PR if you choose to proceed.)

docs/how-to/itemscontrol-how-to.md
```diff
@@ -64,17 +63,21 @@ public class MainViewModel
 
 ### Custom panel
 
-`ItemsControl` displays content in an `ItemsPanel`, which is a `StackPanel` by default. To change how items are arranged, you can replace it with a different control. Use `ItemsPanelTemplate` to override the default, then add a layout control of your choice.
+`ItemsControl` displays its content in an `ItemsPanel`, which is a [layout control](/controls/) that defaults to `StackPanel`.
```
MrJul: The added link lists all controls, not only panels, which are the only one usable in an `ItemsPanel`. Not sure if that's expected.

docs/how-to/itemscontrol-how-to.md
```diff
@@ -123,26 +127,30 @@ public class MainViewModel
...
+    <ItemsControl.ItemsPanel>
+      <ItemsPanelTemplate>
+        <StackPanel Orientation="Horizontal" Spacing="16" />
+      </ItemsPanelTemplate>
+    </ItemsControl.ItemsPanel>
+    <ItemsControl.ItemTemplate>
+      <DataTemplate>
+        <StackPanel Width="120">
+          <Border Width="40" Height="40" CornerRadius="20"
+                  Background="#6366F1" HorizontalAlignment="Center">
```
MrJul: Nit: we might as well reformat those attributes to one per line, matching the format below.

docs/how-to/itemscontrol-how-to.md
```diff
@@ -175,140 +183,106 @@ public class MainViewModel
...
-                <Border Padding="8" Margin="0,0,0,4"
-                        BorderBrush="#E0E0E0" BorderThickness="0,0,0,1">
-                    <TextBlock Text="{Binding Title}" />
-                </Border>
-            </DataTemplate>
-        </ItemsRepeater.ItemTemplate>
-    </ItemsRepeater>
-</ScrollViewer>
-```
+`ItemsControl` does not include scrolling by default. If you wish to display items in a scrollable but non-selectable field, wrap an `ItemsControl` inside a `ScrollViewer`.
```
MrJul: It's already been mentioned, but it might be worth saying again that a `ListBox` is probably a better choice. I think adding a sample showing a virtualized `ItemsControl` I mentioned earlier would be valuable.

docs/layout/choosing-a-layout-panel.md
```diff
@@ -288,7 +288,7 @@ For complex layouts, you can combine multiple panels. Nest panels within one ano
 
 - Prefer simpler panels when possible. [`StackPanel`](#stackpanel) and [`Panel`](#panel) are lighter than [`Grid`](#grid).
 - Avoid deeply nested panels. If you find yourself nesting more than three levels deep, consider whether a single [`Grid`](#grid) with the right row and column definitions could replace the entire tree.
-- For large scrollable lists, use [`ItemsRepeater`](/controls/data-display/collections/itemsrepeater) or [`ListBox`](/controls/data-display/collections/listbox) with virtualization instead of putting hundreds of controls in a `StackPanel` inside a `ScrollViewer`.
+- For lists with many items, use `VirtualizingStackPanel` or `ListBox` instead of placing many controls in a `StackPanel`.
```
MrJul: Same remark as before: `VirtualizingStackPanel` isn't usable on its own.

MrJul (changes requested):

controls/data-display/collections/itemscontrol.md
luke-whos-here: I'll split the ScrollViewer part back into a separate bullet point.

controls/data-display/collections/itemsrepeater.md
luke-whos-here: I can see the argument for that approach, but I think it's better not to. Main reason being the page is only indexed if it's in the sidebar, and it's not a great user experience to have random "Nothing to see here" pages cluttering the navigation.

controls/layout/panels/stackpanel.md
luke-whos-here: Makes sense. In that case, might be easier just to remove the mention of VirtualizingStackPanel.

docs/layout/choosing-a-layout-panel.md
luke-whos-here: Will do the same and remove VirtualizingStackPanel from the sentence.

docs/how-to/itemscontrol-how-to.md
luke-whos-here: I wish I'd thought of that earlier!

docs/how-to/itemscontrol-how-to.md
luke-whos-here: It's because we no longer have an index page for each category of control. You're right though, the link doesn't really seem to connect well.

docs/how-to/itemscontrol-how-to.md
luke-whos-here: I'll implement the same for all the samples on this page.

docs/how-to/itemscontrol-how-to.md
luke-whos-here: I agree, this is a good idea.

luke-whos-here:
@MrJul Thanks for the detailed review. I've made some further changes that should resolve the feedback you raised.

MrJul (approved):
LGTM!

